PR #20, opened 2022-01-14
updated reports so that they can be sent to credoai 

IanAtCredo:
## Change description
Added report integration to send to credo.

Before this is merged should also fix the seemingly broken metric integration.


credoai/integration.py
```diff
@@ -71,21 +67,20 @@ class Metric(Record):
     def __init__(self,
```
IanAtCredo, line 67: @aproxacs , changes so far. Will add the "metric key" when you tell me what that is

credoai/integration.py
```diff
@@ -71,21 +67,21 @@ class Metric(Record):
...
         self.value = value
-        self.dataset_label = dataset_label
+        self.dataset_variant = dataset_variant
         self.config_hash = dict_hash({k:v for k,v in self.__dict__.items() 
                                       if k!='value'})
     
     def _struct(self):
         return {
-            'metric_id': self.config_hash,
+            'key': self.config_hash,
```
IanAtCredo, line 81: @aproxacs changes to the metric integration

credo-eli (approved):

credoai/integration.py
```diff
@@ -71,21 +67,21 @@ class Metric(Record):
...
-            dataset_label,
+            dataset_variant = None,
             **metadata):
         super().__init__('metrics', **metadata)
         self.metric_type = metric_type
         self.value = value
-        self.dataset_label = dataset_label
+        self.dataset_variant = dataset_variant
         self.config_hash = dict_hash({k:v for k,v in self.__dict__.items() 
                                       if k!='value'})
```
aproxacs, line 77: This hash value will be always different because of creation_time. I think it should be restricted to the several attributes like metric_type, name, and dataset_variant_id.

Also name is highly recommended to be included. The web app's metric view will be confusing without the name.

credoai/integration.py
```diff
@@ -71,21 +67,21 @@ class Metric(Record):
...
         return {
-            'metric_id': self.config_hash,
+            'key': self.config_hash,
             'type': self.metric_type,
             'value': self.value,
-            'dataset': self.dataset_label,
+            'dataset_variant_id': self.dataset_variant,
             'metadata': {'creation_time': self.creation_time,
                         **self.metadata},
             '$type': 'model_metrics'
```
aproxacs, line 87: Please add `updated_at` attribute as well. Sever will use this time as a metric updated time.(It will be implemented soon)
```
'updated_at': self.creation_time.
```

credoai/integration.py
IanAtCredo: Nice catch - I'll remove `creation time`. And I'll have a name always supplied.

PR #26, opened 2022-01-21
NLP Generator Upgrade

amrasekh:
* Enabled multiple models run support and more customizability
* Added prompts datasets with standard schema
* Enabled reporting
* Updated demo and documentations


credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+
+    This module assesses language generation models based on various prompts and assessment attributes
+
+    Parameters
+    ----------
+    generation_config : dict
+        generation models configuration dictionary where keys are names of the models and values are their callable generation functions (input : str, output: str)
+        Example: my_generation_config = {'gpt1': gpt1_text_generator, 'gpt2': gpt2_text_generator}
+
+    assessment_config : dict
```
IanAtCredo, line 33: This whole `assessment_config` confuses me. These seem like they should be arguments for initializing the NLPGeneratorAnalyzer, and are somewhat needlessly buried in a separate dict, making documentation difficult.

credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+
+        dfrunst = dfruns[
+            dfruns["response"] != "nlp generator error"
+        ].copy()  # exclude cases where generator failed to generate a response
+
+        dfrunst_assess = pd.DataFrame(
+            data=None, columns=list(dfrunst.columns) + ["value", "assessment_attribute"]
+        )
+
+        if "assess_with_builtin_models" not in self.assessment_config:
```
IanAtCredo, line 192: why not provide the functionality to both use a build-in-model AND there own functions? Wouldn't they all just end up being different assessments of the resultant generators?

credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+                "The prompts dataset cannot be loaded. Ensure the provided prompt_dataset value is either a path to a valid csv file"
+                + " or name of one of the builtin datasets (i.e."
+                + builtin_prompts_names
+                + "). You provided "
+                + prompt_dataset
+            )
+
+        return df
+
+    def build_perspective_client(self, api_key):
```
IanAtCredo, line 290: This (and other perspective functions) look like they should start with "_"

IanAtCredo (changes requested):
I've left a number of comments on some things that should be changed.

Broadly, there are stylistic things, functional things, and usability things.

**Stylistic things**
* Areas where the code is repeated. 
* Private and public methods are randomly ordered. I propose that we order our methods like so:  [magic methods](https://rszalski.github.io/magicmethods/), public methods, private methods. Talked about [here](https://stackoverflow.com/questions/10289461/what-is-a-good-way-to-order-methods-in-a-python-class/36967648). We should be standard about it.

**Functional Things**
* Reporting needs to be separated
* Using prebuilt assessment models doesn't seem logically incompatible with passing assessment functions. We should just run both

**Usability Things**
* The assessment_config is confusing to me. I think it should be split into arguments to the function
* The `prepare_results` function complains if you haven't run the assessment yet. Why not just fall back on doing that? (I don't have a very strong opinion here, just wondering why you chose not to do that).
* Results were kept in `self.raw_results` and not returned. I changed that as I expected results to be returned.
* Maybe have a `get` function for the results. We could make this a standard part of modules - all store results and a get function accesses them. More imoprtant for heavy "run" functions like here, but we could just make it standard. What do you think?

IanAtCredo:
We should remove all `append` uses. It is deprecated:

image

And is there a faster way to run perspective api? You mention run with fewer iterations - why don't we set that by default? Is there a benefit for multiple iterations that's important for the demo? What would be a smaller prompt dataset?

IanAtCredo:
Ok, standardized everything. 
* Now `run` always creates a `self.results` objects and returns `self`
* `prerpare_results` potentially transforms self.results and returns. If self.results doesn't exist a `NotRunError` is raised
* `get_results` returns self.results. If self.results is None a `NotRunError` is raised.

credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+                .agg(mean=("value", "mean"), std=("value", "std"))
+            )
+            results.sort_values(
+                by=["generation_model", "assessment_attribute", "group"], inplace=True
+            )
+            results = results[
+                ["generation_model", "assessment_attribute", "group", "mean", "std"]
+            ]
+            return results
+        else:
```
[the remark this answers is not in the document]
amrasekh: Resolved per our conversation.

credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+            "languages": ["en"],
+        }
+        response = (
+            self.perspective_client.comments().analyze(body=analyze_request).execute()
+        )
+        return response["attributeScores"][assessment_attribute]["summaryScore"][
+            "value"
+        ]
+
+    def create_reports(self, save_dir=None):
```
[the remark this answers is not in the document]
amrasekh: This is now removed and moved to reporting as an standalone class `NLPGeneratorAnalyzerReport`

credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+                        + str(n_iterations)
+                        + " with Generation Model "
+                        + gen_name
+                    )
+                df["response"] = df["prompt"].apply(
+                    lambda x: self._gen_fun_robust(x, gen_fun)
+                )
+                df["run"] = i + 1
+                df["generation_model"] = gen_name
+                dfruns = dfruns.append(df)
```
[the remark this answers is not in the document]
amrasekh: now replaced with saving all to a list followed by a `concat`

credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+
+                fpath = os.path.join(save_dir, "lens_comprehensive_table_" + dtstr + ".csv")
+                self.raw_results.to_csv(fpath, index=False)
+
+        else:
+            raise Exception(
+                "This NLPGeneratorAnalyzer instance is not run yet. Call 'run' with appropriate arguments before using this module."
+            )
+
+    def perform_prerun_checks(self):
```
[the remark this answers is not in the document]
amrasekh: `run` is now much more concise

credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+class NLPGeneratorAnalyzer(CredoModule):
+    """
+    NLP generation analyzer for Credo AI
+
+    This module assesses language generation models based on various prompts and assessment attributes
+
+    Parameters
+    ----------
+    generation_config : dict
+        generation models configuration dictionary where keys are names of the models and values are their callable generation functions (input : str, output: str)
```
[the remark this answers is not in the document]
amrasekh: made more concise now

credoai/modules/model_assessments/nlp_generator.py
```diff
@@ -0,0 +1,496 @@
...
+from absl import logging
+from credoai.data.utils import get_data_path
+from credoai.modules.credo_module import CredoModule
+from credoai.reporting import plot_utils
+from credoai.utils.common import wrap_list
+from datetime import datetime
+from googleapiclient import discovery
+from time import sleep
+
+matplotlib.rcParams.update({"font.size": 12})
```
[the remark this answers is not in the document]
amrasekh: Removed now.

credoai/modules/model_assessments/nlp_generator.py
amrasekh: updated now to more clear design
```
        prompts,
        generation_functions,
        assessment_functions,
        perspective_config=None,
```

credoai/modules/model_assessments/nlp_generator.py
amrasekh: now enabled. `assessment_functions` parameter can take both types simultaneously

credoai/modules/model_assessments/nlp_generator.py
amrasekh: private function now are all updated to reflect this

credoai/modules/model_modules/nlp_generator.py
```diff
@@ -0,0 +1,372 @@
...
+                    "'generation_functions' and 'assessment_functions' values must be of type dict."
+                )
+
+        # Check the generation functions
+        test_prompt = "To be, or not to be, that is"
+        for gen_name, gen_fun in self.generation_functions.items():
+            try:
+                response = gen_fun(test_prompt)
+                if not isinstance(response, str):
+                    raise Exception(
```
IanAtCredo, line 321: All of these exceptions should be made `ValidationErrors` (see credoai.utils.common). I am making some errors we can return that are more informative. This prerun_check is definitely a form of validation

IanAtCredo:
@amrasekh , I have brought the NLP Generator into Lens. I had to make some decisions about the default prompt set to use and set up a default generator function (for gpt2) to be the comparison class. Also had to define the default assessment function. Most of that is included in `credoai.assessments.nlp_utils.py`

Added a `create_reports` function to the assessment as you can see. This allows lens to run no problemo.

Overall, very easy to make an assessment (though some decisions had to be made) once the module was made.

Corrected a bug with 'perspective_config' (you checked for the string instead of self.perspective_config) and moved over to validation errors.

IanAtCredo (approved):
I've tested quite a bit, but I'm sure there are still things that can be improved and may be broken.

Please test the lens additions before merging!

amrasekh:
> We should remove all `append` uses. It is deprecated:
> 
> image
> 
> And is there a faster way to run perspective api? You mention run with fewer iterations - why don't we set that by default? Is there a benefit for multiple iterations that's important for the demo? What would be a smaller prompt dataset?

The `append` s have been replaced with `concat`.
I changed the number of iterations to 1. The `bold_religious_ideology` is the smallest dataset with protected attributes that we have builtin now, so that one cannot be made changed. I'll create and add a small counterfactual dataset based on [this](https://raw.githubusercontent.com/conversationai/unintended-ml-bias-analysis/main/sentence_templates/en_sentence_templates.csv) later.

credoai/modules/model_modules/nlp_generator.py
amrasekh: I see you have now made the updates accordingly. Awesome.

PR #32, opened 2022-01-26
Ian/log models

IanAtCredo:
## Change description
Model registration from within lens


credoai/integration.py
```diff
@@ -86,7 +86,6 @@ def _struct(self):
             'name': self.name,
             'value': self.value,
             'process': self.process,
-            'value_updated_at': self.creation_time,
```
IanAtCredo, line 89: @aproxacs , removed `value_updated_at` as it seemed to cause problems showing metrics.

aproxacs: @IanAtCredo  `value_updated_at` should be IOS format like `2022-01-27T13:17:19.875120`
Please use `self.creation_time.isoformat()`

aproxacs: Oh, creation time is string already.  
```
self.creation_time = datetime.now().strftime("%d-%m-%Y %H:%M:%S")
```
Why don't we keep this as datetime?
```
self.creation_time = datetime.now()
```
And change it as string when converting to json?

aproxacs:
@IanAtCredo I think we talked about using the same format for both of json file upload and API data. Then `jsonify()` and `credoify()` should be the same. 

This PR of credo-ui(https://github.com/credo-ai/credo-ui/pull/302) has applied the changes. 

Let me know if I am missing something.

credoai/integration.py
IanAtCredo: I can do that, sure. But just so you know, the metrics uploaded are NOT reflecting the `value_updated_at` I am sending

IanAtCredo:
> @IanAtCredo I think we talked about using the same format for both of json file upload and API data. Then `jsonify()` and `credoify()` should be the same.
> 
> This PR of credo-ui([credo-ai/credo-ui#302](https://github.com/credo-ai/credo-ui/pull/302)) has applied the changes.
> 
> Let me know if I am missing something.

We did? Ok, I'll make `jsonify` an alias for `credoify`

credoai/integration.py
aproxacs: @IanAtCredo This PR(https://github.com/credo-ai/credo-ui/pull/302) is handling 'value_updated_at'. But it has not yet been deployed to QA

IanAtCredo: Got it 👍

aproxacs (approved):
Thank you!

PR #39, opened 2022-02-04
Ian/fix import clean

IanAtCredo:
## Change description
This PR should fix the issues with import NLPGenerator assessment. 

I have uploaded a new version to testpypi (install as so: 
`pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple credoai-lens` ). Please install into a clean environment and test.

Changes:
* nlp_utils has been moved to utils. Unless you explicitly import it `import credoai.utils.nlp_utils` it won't lead to failure.
* The `NLPGenerator` assessment now uses `cutils.nlp_utils` in its `init_module` function. That will fail if somehow initialized without the correct dependencies, but it will raise an informative error.
* Introduced "installationError" to use occasionaly.

Intended behavior:
* Because NLPGenerator module also requires extras to be installed, NLPGenerator assessment will not be accessible unless the requirements are installed

Caveat:
* If there is a future where a module has not additional requirements, but the _assessment_ does for some reason, then we will be in a bad way. We should strive for the assessments to not have additional dependencies beyond the relevant modules.

Other notes:
* We don't have clear delineation between things that require extras and those that don't. I have started labeling files for things that require extras.


amrasekh (approved):
The issues seem to have been resolved now with this. 
Note: merging this would most probably create some conflicts for the other two other PRs, but please merge (as it is higher priority) and I take care of those conflicts afterwards.

PR #41, opened 2022-02-04
Documentations update

amrasekh:
Added and updated docstring, rst, and notebook documentations


docs/tutorials.rst
```diff
@@ -7,12 +7,52 @@ Tutorials
...
    notebooks/integration_demo
+   notebooks/lens_customization
 
 
 .. toctree::
    :caption: Demo Use Cases
    :maxdepth: 1
 
    notebooks/lens_demos/binaryclassification
+   docs/notebooks/module_demos/fairness_nlp
```
IanAtCredo, line 19: this should be `notebooks/module_demos...` not `docs/notebooks/module_demos...`

However, I don't think we should put `module_demos` here. We should have another section for them. Using modules directly should be hidden as much as possible (that's the current working hypothesis anyway). I want to direct people to use lens.

credoai/assessment/nlp_utils.py
```diff
@@ -11,6 +23,14 @@ def gpt2_text_generator(prompt):
     return response
 
 def get_default_nlp_assessments():
+    """Loads the default nlp assessment model
+    The model assesses a text for toxicity
```
IanAtCredo, line 27: missing new line

IanAtCredo (approved):
Minor changes requested before you merge.

credoai/assessment/nlp_utils.py
amrasekh: added now

docs/tutorials.rst
amrasekh: Agreed. Need to please both people who are more into exploring repos and others who prefer readthedocs.

amrasekh: link fixed

PR #48, opened 2022-02-08
updated notebook descriptions in readthedocs, added lens demo for nlp…

IanAtCredo:
… generator, added more documentation about assessments

## Change description

Few things here. This PR adds more to finalize the NLP assessment. It fixes some issues with the assessment, adds documentation, creates a demo for it, etc. @amrasekh, check this out to see how _far_ the NLP assessment was from being "done". It wasn't a ton of work, but still a big difference from finishing the module alone. 

In general, I think it is better to work on creating a `lens demo` when we add functionality. We don't really need more `module demos`. If we have them, great, but not critical. 

Reorganized and simplified the notebook descriptions in readthedocs. Added headers to make it clearer. Separated out module demos and explicitly deprioritized them


docs/notebooks/quickstart.ipynb
```diff
@@ -432,6 +432,85 @@
...
+   "source": [
+    "#### Assessment Documentation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "db8ed40f-578a-4586-b0a5-28c680db7686",
+   "metadata": {},
+   "source": [
+    "Documentation for each assessment is a bit difficult to understand. Assessments wrap modules, which themselves have documentation. Normally, you don't have to worry about the module itself, except if you are creating your own assessments or want to use the modules directly. The main exception here is if you want to parameterize how you run the assessment\n",
```
IanAtCredo, line 448: Calling it explicitly that documentation for each assessment is difficult to get. I think removing the `init_module` function may help here. But it's still a bit confusing.

credoai/lens.py
```diff
@@ -599,10 +600,13 @@ def _get_names(self):
     def _init_assessments(self):
         """Initializes modules in each assessment"""
         for assessment in self.assessments.values():
-            kwargs = self.spec.get(assessment.name, {})
-            assessment.init_module(model=self.model,
-                                   data=self.data,
-                                   **kwargs)
+            kwargs = deepcopy(self.spec.get(assessment.name, {}))
```
IanAtCredo, line 603: @amrasekh , this is how I am using the requirements of an assessment to determine if it requires "model" or "data".  I can imagine us doing soemthing similar when deciding whether to export to "model" or "data"

PR #54, opened 2022-02-10
Data assess reporting

amrasekh:
Created dataset assessment visualization and reporting and embedded in Lens


credoai/reporting/dataset_base.py
```diff
@@ -0,0 +1,185 @@
...
+from credoai.reporting.credo_report import CredoReport
+from credoai.reporting import plot_utils
+
+
+class DatasetModuleReport(CredoReport):
+    def __init__(self, module):
+        super().__init__()
+        self.module = module
+
+    def create_report(self, filename=None):
```
IanAtCredo, line 14: Please separate this into separate functions for each plot

credoai/reporting/dataset_base.py
```diff
@@ -0,0 +1,185 @@
...
+            x="count",
+            y=sensitive_feature_name,
+            hue=label_name,
+            data=df,
+            palette=plot_utils.credo_diverging_palette(num_classes),
+            alpha=1,
+            ax=axs[1],
+        )
+        fig.patch.set_facecolor("white")
+        ax.set_frame_on(False)
```
IanAtCredo: I've changed these to `sns.despine`. In general I think it's better (not always) to have the axis lines on the left and bottom

credoai/reporting/dataset_base.py
```diff
@@ -0,0 +1,186 @@
...
+        ax = sns.barplot(
+            x="count",
+            y=sensitive_feature_name,
+            data=df,
+            palette=plot_utils.credo_diverging_palette(1),
+            alpha=1,
+            ax=axs[0],
+        )
+        fig.patch.set_facecolor("white")
+        sns.despine()
```
IanAtCredo, line 45: I changed all `frame_off` to `despine`, which keeps the left and bottom axes' lines. Think it looks better

credoai/reporting/dataset_base.py
```diff
@@ -0,0 +1,186 @@
...
+        filename : string, optional
+            If given, the location where the generated pdf report will be saved, by default Non
+
+        Returns
+        -------
+        array of figures
+        """
+        results_all = self.module.get_results()
+
+        # Generate sample balance barplots
```
IanAtCredo, line 28: This and the label balance plot could be made smaller and brought together. I'd suggest one plot with two subplots.

credoai/reporting/dataset_base.py
```diff
@@ -0,0 +1,186 @@
...
+            ax.axhline(0, color='k')
+            sns.despine()
+            plt.title("Group differences for " + k + " combination across features")
+            plt.xlabel("")
+            plt.ylabel("Group difference")
+            plt.xticks(rotation=90)
+
+            self.figs.append(fig)
+
+        # Generate mutual information barplots
```
IanAtCredo, line 141: Nit: For some reason the categorical and continuous barplots are not centered and offset in different directions

image

IanAtCredo (changes requested):
High level comments:
* Overall proxy score should be shown. Probably just as a number.
* Labels should be made prettier - e.g., no underscore. See `credoai.reporting.plot_utils.format_metric` 
     * Probably want to change the name of that function to `format_label` or something
     * Not sure if we want to want to mess with the case of the label. I think that caused issues in the past. But all upper case looks weird.
     
General thoughts about reports (not to be addressed in this PR!)
* Our reports are currently a bit confusing. Could use some text, better formating, etc. I'd love to spend our time in person whiteboarding ideal reports for fairness, nlp and datasets.
* Having parity and ratio metrics on the same plot leads to problems. Probably want to know more about those metrics

credoai/reporting/dataset_base.py
```diff
@@ -0,0 +1,186 @@
...
+        plt.tight_layout()
+
+        self.figs.append(fig)
+
+        # Generate group difference barplots
+        results = results_all["group_diffs"]
+        for k, v in results.items():
+            df = pd.DataFrame(v.items(), columns=["feature", "group difference"])
+            fig = plt.figure(figsize=(8, 4), dpi=150)
+            ax = sns.barplot(
```
IanAtCredo, line 124: Rather than create a number of different plots, create one figure with the correct number of subplots (the length of results.items()) and fill them in.

This lets us potentially create a grid of subplots, rather than just one after the other. And add a supertitle, control overal aesthetics, ec.

credoai/reporting/dataset_base.py
amrasekh: They are together in the same figure as subplots. The demographic parity metrics is also there because it's also about data balance (immediately calculated from these two figures actually). I made some style updates.

credoai/reporting/dataset_base.py
amrasekh: Fixed now.

credoai/reporting/dataset_base.py
```diff
@@ -0,0 +1,186 @@
...
+            hue="feature type",
+            data=df,
+            palette=plot_utils.credo_diverging_palette(num_types),
+            alpha=1,
+        )
+        fig.patch.set_facecolor("white")
+        sns.despine()
+        plt.title("Mututal information with " + ref_type + " feature " + ref_name)
+        plt.xlabel("")
+        plt.ylabel("Mutual information")
```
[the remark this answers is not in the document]
amrasekh: Applied. ylim is also now 0-100%.

amrasekh:
I don't quite know how to visualize the overall proxy score as it is just one number and also adding it to any of the current plots could become misleading. Seems to better be accommodated through extending pdf generators to display text/tables too.

IanAtCredo:
I'd put it as one large number. I think standard text can work for this. But I think we can talk about this within the broader context of "executive summary" section of our reports - with big numbers reflecting things and won't block this merge

credoai/reporting/dataset_base.py
```diff
@@ -23,11 +24,34 @@ def create_report(self, filename=None):
         -------
         array of figures
         """
-        results_all = self.module.get_results()
+        self.results_all = self.module.get_results()
```
IanAtCredo, line 27: Nit: not best practice to define instance variables outside of __init__. In the NLP reports I just get the results in each plot function. You can also just define this in the __init__

credoai/reporting/dataset_base.py
```diff
@@ -159,7 +199,7 @@ def create_report(self, filename=None):
         ref_name = ref.iloc[0]["feature"]
         ref_type = ref.iloc[0]["feature type"].split("_")[0]
 
-        fig = plt.figure(figsize=(8, 4), dpi=150)
+        fig = plt.figure(dpi=150)
```
IanAtCredo, line 202: I saw you setting dpi, and thought that made sense. So I added a "get_style" function that I started using in the nlp report. We can standardize our plotting later. See [here](https://github.com/credo-ai/credoai_lens/blob/ian/nlp_report_improv/credoai/reporting/nlp_generator.py#L51)

credoai/reporting/dataset_base.py
```diff
@@ -116,30 +140,46 @@ def create_report(self, filename=None):
...
             df = pd.DataFrame(v.items(), columns=["feature", "group difference"])
-            fig = plt.figure(figsize=(8, 4), dpi=150)
-            ax = sns.barplot(
-                x="feature",
-                y="group difference",
-                data=df,
-                palette=plot_utils.credo_diverging_palette(1),
-                alpha=1,
-            )
+            if len(results) > 1:
```
IanAtCredo, line 151: This works even if len(results) == 1 - no need for conditional logic

IanAtCredo (approved):

credoai/reporting/dataset_base.py
amrasekh: That's also what I though but threw out `TypeError: 'AxesSubplot' object is not subscriptable`. Tried to fixed it with parameters, but not success. So opted for this solution.

credoai/reporting/dataset_base.py
amrasekh: Good idea. Applied.

PR #65, opened 2022-02-25
added visualization updates for all assessments. Split out from PR #63

IanAtCredo:
## Change description

Splits out visualization updates from #63.


credoai/reporting/dataset_base.py
```diff
@@ -47,134 +45,116 @@ def _plot_balance_metrics(self):
...
+                bbox_to_anchor=(1.2, 0.5), 
+                loc="center",
+                frameon=False,
+                ncol=num_classes,
+                title=label_name
+            )
+            ax.legend_.set_title(label_name)
+        self.figs.append(f)
 
     def _plot_group_diff(self):
```
amrasekh, line 132: This and the mutual information plots, which are both bar plots and would look better in landscape, are now plotted in portrait aspect ratio. 
image

credoai/reporting/plot_utils.py
```diff
@@ -6,12 +6,62 @@
 
 DEFAULT_COLOR = '#4937c0'
 
-def get_style(rc=None, figsize=6, figure_ratio=1):
-    figsize = [figsize, figsize*figure_ratio]
-    style = {'figure.figsize': figsize,
-             'figure.dpi': 200}
```
amrasekh, line 12: For some reason, this `dpi` setting is not actually applied in some of the plots. If quickstart demo is run, this for example is some plots are made.
image
image

amrasekh (approved):
Included some suggestions/issues about the aspect ratio, text layout, and resolution of some of the plotting functions.

credoai/reporting/plot_utils.py
IanAtCredo: I'll look into this.

credoai/reporting/dataset_base.py
IanAtCredo: Good point - fixed

credoai/reporting/plot_utils.py
IanAtCredo: Issue I found, which I have fixed.
* Jupyter defaults to displaying plots as png's.
*  It uses dpi as a scaling factor. So it isn't that easy to create a plot of a specific size and up the resolution.
*  I can make it so the plots are automatically created as svg's. This solves the issues (with some minor overhead for plotting). I've done this by adding the lines to all of our demo notebooks (and later to our reports).

PR #66, opened 2022-02-28
Nl gen context

amrasekh:
Contextualization of NLP Generator Assessment Results
Provides context for Perspective API text attribute intensities
- Shows examples of real-world texts associated with various Perspective API text attribute value levels
- Create benchmark plots for certain text sources (currently includes tweets from @realdonaldtrump, @barackobama, and a toxic user persona)
- Text attributes include: 'PROFANITY', 'TOXICITY', 'IDENTITY_ATTACK', 'INSULT', 'THREAT'


IanAtCredo:
I'd make the plots at the end smaller. It looks like you are using an old usage of `get_style`. See [here](https://github.com/credo-ai/credoai_lens/blob/develop/credoai/reporting/nlp_generator.py#L79) for more up-to-date usage. Now "size" and "figure_ratio" more or less refers to the size of one plot. You can then set n_cols and n_rows and the size will expand.

Is there a short sentence definition for each text attribute?

Censor curse words would be good!

@credo-eli, do we want this kind of "contextualization" work in Lens?

IanAtCredo (approved):
No blocking comments, but a few things I'd want to improve. Let's make sure we want this part of Lens and not part of the app.

credo-eli (approved):

amrasekh:
moved to credoai_research repo

PR #74, opened 2022-03-09
Ian/metric constant restructure

IanAtCredo:
## Change description
Metrics have been reorganized and restructured.

* Metric class introduced. This class extends the functionality of "FairnessFunction" and serves as a common structure to define metrics, their category, their equivalent names, etc.
* some notebooks updates, to make them work with new metric class. some of those notebook were outright broken before.
* Minor bugs that were introduced in previous PRs have been removed.


credoai/metrics/metrics.py
```diff
@@ -0,0 +1,140 @@
+from absl import logging
```
IanAtCredo, line 1: This is the main new functionality. Most of the other changes are dealing with this change.

amrasekh (approved):

PR #76, opened 2022-03-10
Ian/regression reporter

IanAtCredo:
## Change description
This builds on #74.

Additions:
* Created FairnessReporter and BinaryClassificationReporter (inherits from FairnessReporter)
* FairnessReporter drops the infographics for more flexibility

Bug fixes:
* registering model to use case was failing - not updated with new API
* custom metrics were failing with new reporting. Couldn't pickle! Cloudpickle solves this issue.

Pickling issues in more detail. If you had the follow custom metric:
```
def custom_fun(y_true, y_pred):
     return foo

custom_metric = Metric('my_metric', fun=custom_fun)
```
And tried to pickle "custom_metric", unpickling would fail in a new environment because "custom_fun" isn't in the namespace. This comes up in report generation because we rely on pickling/unpickling the reporter to transfer information from the main code to the generated report notebook.

The package `Dill` succeeds in unpickling, but it doesn't actually work. It just doesn't complain about not having `custom_fun`. Cloudpickle solves this issue through blackmagic.


credoai/reporting/reports.py
```diff
@@ -91,17 +97,17 @@ def __init__(self, needed_artifacts=None):
         super().__init__()
         self.needed_artifacts = needed_artifacts
         # set up reporter
-        load_code="import pickle\n"
+        load_code="import cloudpickle\n"
```
IanAtCredo, line 100: Moving stuff over to cloudpickle

credoai/utils/credo_api_utils.py
```diff
@@ -243,20 +247,11 @@ def register_project(project_name):
...
-        model_ids.append(model_id)
-    else:
-        return
-    data = serialize({'model_ids': model_ids, 
-                      'id': 'resource-id',
-                      '$type': 'use_case_id_scopes'})
-    end_point = get_end_point(f"use_cases/{use_case_id}/scopes/draft")
-    submit_request('patch', end_point, data=json.dumps(data), headers={"content-type": "application/vnd.api+json"})
     
+def register_model_to_use_case(use_case_id, model_id):
```
IanAtCredo, line 251: Update to model registration to conform to new API

credoai/reporting/model_fairness.py
```diff
@@ -124,6 +123,110 @@ def plot_fairness(self):
...
+    def _style_barplot(self, ax):
+        sns.despine()
+        ax.set_xlabel(ax.get_xlabel(), fontweight='bold')
+        ax.set_ylabel(ax.get_ylabel(), fontweight='bold')
+        # format metric labels
+        ax.set_yticklabels([format_label(label.get_text()) 
+                            for label in ax.get_yticklabels()])
+    
+
+class BinaryClassificationReporter(FairnessReporter):
```
IanAtCredo, line 175: Made a child of FairnessReporter that separates the more specific BinaryClassification functionality from the general functionality

PR #81, opened 2022-03-11
Ian/performance

IanAtCredo:
## Change description

###Added PerformanceModule
Performance module instantiates a subset of fairnessmodule's functionality. Specifically, calculating overall score and disaggregated performance. If no sensitive_feature is provided, no disaggregated performance is calculated. Fairness module no longer calculates overall scores.

Performance module does not _require_ a sensitive feature, while fairness module does.

If both performance module and fairness module are used, we will get two disaggregated performance values. E.g., there is overlapping functionality. Not sure how to deal with this now.

CredoData now doesn't require a sensitive feature.


credoai/modules/model_modules/performance_base.py
```diff
@@ -0,0 +1,266 @@
...
+from credoai.modules.credo_module import CredoModule
+from fairlearn.metrics import MetricFrame
+from scipy.stats import norm
+from sklearn.utils import check_consistent_length
+from typing import List, Union
+import pandas as pd
+
+class PerformanceModule(CredoModule):
+    """
+    Fairness module for Credo AI. Handles any metric that can be
```
amrasekh, line 13: "Fairness module"? Seems like it should be Performance module

amrasekh (approved):
Looks good. I suggest adding a demo too for it in this PR already.

IanAtCredo:
> Looks good. I suggest adding a demo too for it in this PR already.

I don't think we should delay new features based on new modules. But I realize quickstart will fail if we add new modules but don't update the assessment spec.

PR #101, opened 2022-04-13
Feat/metrics wo lens

IanAtCredo:
## Change description
Moved export functionality to CredoGovernance. This allows the governance object to be created and used to export metrics that may be calculated by a system outside of Lens.

Simplified Lens export code to now make use of introduced functionality. 

Minor bug fixes along the way.


amrasekh (approved):
Possibly this feature can also be showcased/demoed in the [governance demo](https://github.com/credo-ai/credoai_lens/blob/develop/docs/notebooks/governance_integration.ipynb) demo as well.

PR #110, opened 2022-04-22
Metrics page

amrasekh:
Created the metrics page for Lens documentations.


IanAtCredo (changes requested):
So this page is completely manual - is there no way to create this page automatically? If not, ok.

I don't believe this page is being shown in documentation anywhere yet. I added a build for this branch to our [readthedocs](https://readthedocs.org/projects/credoai-lens/versions/), which is how you can see what's going on (if you don't want to build locally)

IanAtCredo:
Check [this page](https://stackoverflow.com/questions/50831070/how-to-document-linkable-constants-with-sphinx). I think we may be able to set this up so we can automatically reference the constants

amrasekh:
The metrics rst page is now added to the index.
A next-step option for automation is to embed this metrics.rst in the `metrics_constants.py` and setup up a process to pull the info from there for documentation.

IanAtCredo (approved):

PR #114, opened 2022-04-27
Added percentage labels to data balance plot

amrasekh:
A small PR that add percentage labels to data balance plot.
Runningun [dataset_assessment.ipynb](https://github.com/credo-ai/credoai_lens/blob/develop/docs/notebooks/lens_demos/dataset_assessment.ipynb):

image
image

### DIFF
--- a/credoai/lens.py
+++ b/credoai/lens.py
@@ -267,6 +267,7 @@ class Lens:
                 continue
             reporter.display_results_tables()
             reporter.plot_results()
+        return self
 
     def _apply_dev_mode(self, dev_mode):
         if dev_mode:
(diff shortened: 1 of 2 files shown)


IanAtCredo:
Added additional functionality. Now if there is room it will plot in the center of bars (which I think looks better). Also added to all bar plots, which required some patch manipulations!

image

IanAtCredo (approved):
I added some things. Looks good!

amrasekh:
Looks nicer now for SEX feature but for RACE, which has multiple groups, it is not working. All bars in the first plot are also labeled with 100%.
image

IanAtCredo:
I see... looking into this

IanAtCredo:
image
image

Fontsize is the hardest thing. But this looks ok. Would be better to just have percentages on hover!

PR #117, opened 2022-05-03
Metric search exception handling

amrasekh:
This small PR addresses cases where Lens does not find any matches for a user-specified metric name.

### DIFF
--- a/credoai/modules/model_modules/fairness_base.py
+++ b/credoai/modules/model_modules/fairness_base.py
@@ -206,9 +206,9 @@ class FairnessModule(PerformanceModule):
                 if len(metric) == 1:
                     metric = metric[0]
                 elif len(metric) == 0:
-                    raise Exception("Returned no metrics when searching using a metric name. Expected to find one matching metric.")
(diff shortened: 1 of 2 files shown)


IanAtCredo (approved):
This seems fine, but I thought the bug we saw was communicating the "multiple metric" text when it wasn't finding any metric (e.g. "accuracy" wasn't found). Does this correct that?

amrasekh:
> This seems fine, but I thought the bug we saw was communicating the "multiple metric" text when it wasn't finding any metric (e.g. "accuracy" wasn't found). Does this correct that?

Yes.

Before (succeed if one matching metric found, otherwise this error below):
`Exception: Returned multiple metrics when searching using a metric name. Expected to find only one matching metric`

Now (succeed if one matching metric found, the error above if multiple found, the error below if none found): 
`Exception: Returned no metrics when searching using the provided metric name <metric_name>. Expected to find one matching metric.`

Which means for `accuracy` it now throws:
`Exception: Returned no metrics when searching using the provided metric name <accuracy>. Expected to find one matching metric.`

IanAtCredo:
Sounds good. If you think everything has been taken care of, please merge!

PR #120, opened 2022-05-08
moved over from IDs to names

IanAtCredo:
## Change description
We are transitioning from IDs to names in the interface in Lens. This happens by using the CredoGovernance object to hide IDs and handle transitions between names and IDs.

One place where this is an issue is in recording which dataset is associated with each metric. Users are potentially in an air-gapped situation where we cannot get the ID corresponding to a particular dataset name. As a result, we need to associate metrics with `dataset_name` rather than `dataset_id`. @aproxacs , let's meet briefly to discuss this.

### DIFF
--- a/credoai/integration.py
+++ b/credoai/integration.py
@@ -58,8 +58,8 @@ class Metric(Record):
         metric value
     subtype : string, optional
         subtype of metric. Defaults to base
-    dataset_id : str, optional
-        ID of dataset. Should match a dataset on Governance App
+    dataset_name : str, optional
+        Name of dataset. Should match a dataset on Governance App
     process : string, optional
         String reflecting the process used to create the metric. E.g.,
         name of a particular Lens assessment, or link to code.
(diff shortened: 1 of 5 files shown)


credoai/integration.py
```diff
@@ -76,14 +76,14 @@ def __init__(self,
                  metric_type,
                  value,
                  subtype="base",
-                 dataset_id=None,
+                 dataset_name=None,
```
IanAtCredo, line 79: @aproxacs - changing to `dataset_name` instead of `id` as part of move away from ids.

amrasekh (approved):
Works as expected. 
I just suggest reviewing the [gov demo](https://github.com/credo-ai/credoai_lens/blob/b2970ef792bf13e403da7bdedb0ece875e651bc6/docs/notebooks/governance_integration.ipynb) for any possible updates missing, like: 

> Note The below will fail unless you change "use-case-id" to the ID of a use-case actually defined in your Governance app!

PR #121, opened 2022-05-11
added basic computational efficiency inference module

IanAtCredo:
## Change description
Small module that calculates the average time of inference for a batch of 1000. Separated into a separate module (rather than performance) because inference _has_ to happen within the module.

This will connect with new risk issues introduced in this [fast-follow PR](https://github.com/credo-ai/rai-service/pull/36)

Example:
image

### DIFF
--- a/credoai/modules/model_modules/computational_efficiency.py
+++ b/credoai/modules/model_modules/computational_efficiency.py
@@ -52,6 +52,7 @@ class ComputationalEfficiencyModule(CredoModule):
             )
 
     def _inference_efficiency(self):
+        """Estimates time to predict 1000 records"""
         t1 = perf_counter()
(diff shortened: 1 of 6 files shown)


credoai/modules/model_modules/computational_efficiency.py
```diff
@@ -0,0 +1,62 @@
...
+            Occurs if self.run is not called yet to generate the raw assessment results
+        """
+        if self.results is not None:
+            return self.results
+        else:
+            raise NotRunError(
+                "Results not created yet. Call 'run' with appropriate arguments before preparing results"
+            )
+
+    def _inference_efficiency(self):
```
amrasekh, line 54: docstring missing

amrasekh (approved):

PR #124, opened 2022-05-16
Feat/new endpoint

IanAtCredo:
## Change description
This changes how CredoGovernance works completely.

Now a single url (or path to file that represents the same content) is passed to CredoGovernance. From this url you will get:
* the use case, model and dataset ids
* the assessment plan (if defined

This current instantiation _requires_ that a use-case and model have already been registered. We are eliminating the ability for Lens to register the model when not defined (which would require passing a use-case without the model).

While this is a minor reduction in functionality, we don't believe model registration will happen on Lens side often, and it _greatly_ simplifies instruction. Now the Lens user can be completely ignorant of the registration dynamics and everything will work fine.

@kunna , right now the credo-url is expected to look like so: "use_cases/{use_case_id}/models/{model_id}/assessment_spec". Do we ultimately want this? Is it an issue that it exposes what our API looks like? Do we want to provide some kind of "masked" url instead?


credoai/utils/credo_api_utils.py
```diff
@@ -76,18 +76,30 @@ def get_associated_models(use_case_id):
...
-        assessment_plan_id = deserialize(
-            submit_request('get', end_point).json())['id']
+        end_point = get_end_point(assessment_spec_url)
+        downloaded_spec = deserialize(submit_request('get', end_point).json())
+        assessment_spec = {k: v for k,
+                           v in downloaded_spec.items() if '_id' in k}
+    except requests.exceptions.HTTPError:
+        raise IntegrationError("Failed to retrieve assessment spec. Check that the url is correct")
+    try:
+        assessment_plan_id = downloaded_spec['id']
```
IanAtCredo, line 94: @kunna , right now I am using the ID of the spec. If you plan to include an "assessment_plan_id" in the payload, as we discussed, let me know and I'll update this.

IanAtCredo:
@kunna , it would be good to include directions in my tutorial notebook about where you can download and fine these assessment plans. It'd also be good to include an example file to showcase. Let me know when those are prepared.

credoai/utils/credo_api_utils.py
kunna: I updated and push the changes.
* there will be `assessment_plan_id` attribute
* request will respond with validate data even if assessment_plan is not defined

credoai/utils/credo_api_utils.py
```diff
@@ -75,6 +75,20 @@ def get_associated_models(use_case_id):
...
 
+def get_assessment_spec(use_case_id, model_id):
+    try:
+        end_point = get_end_point(
+            f"use_cases/{use_case_id}/models/{model_id}/assessment_spec")
+        assessment_spec = deserialize(submit_request('get', end_point).json())
+        assessment_plan_id = assessment_spec['assessment_plan_id']
+        end_point = get_end_point(
+            f"assessment_plans/{assessment_plan_id}/metrics")
+        assessment_plan =  {'metrics': deserialize(submit_request('get', end_point).json())}
```
kunna: I updated the api response to include assessment_plan and metrics. no need to make an extra request.

amrasekh (approved):

PR #127, opened 2022-05-20
Privacy module

amrasekh:
This introduces Privacy Module, which is based on Adversarial Robustness Toolbox. Currently includes
* rule-based member inference privacy attack
* model-based member inference privacy attack

Next Lens steps:
* Create a demo
* Extend to include more attack approaches from the toolbox


IanAtCredo (changes requested):
Could you add a test that ensures that the privacy module is selected when training data is included?

I made a few changes:
* one bug fix (self.y_assess_train => y_assess_train)
* added a new metric (membership_Inference_attack)
* remove "accuracy" from metric names. I believe that's not important at this point. The attack is _defined_ by using accuracy score. In the future we may allow people to define it in other ways, but it should still be defined by just one metric, so we can just calll it score
* Slightly refactored how the functions => run work
* simplified prepare results

amrasekh:
> 

The new tests are now added. The updates look great.

IanAtCredo (approved):

PR #130, opened 2022-05-26
Feat/new reporting

IanAtCredo:
## Change description

Many changes related to this [technical doc](https://credo-ai.atlassian.net/wiki/spaces/MVP/pages/460390401/Technical+UI-Generated+Risk+Assessments). 

Essentially:
* notebook reports are going away
* charts and html blobs (now called files) will be separately uploaded as a list of assets (charts and files)
* each will be associated with a list of metric_keys

To do this:
* Reporters were changed to return lists of assets
* assessments now set a reporter
* Lens has been changed in a bunch of ways
* Notebook Reporting functionality has been completely removed
* Module demos have been removed from Lens

To do:
* Assets sometimes need to be associated with model/dataset id (probably always should be!). Need to update this. Particularly important when an asset isn't associated with any metric (like dataset profiler


credoai/reporting/dataset_fairness.py
```diff
@@ -205,66 +128,7 @@ def _plot_balance_metrics(self):
...
-
-                <pr>While some differences amongst groups should be expected, large deviations
-                are problematic. One of the main issues is that they may lead
-                to your dataset <em>redundantly encoding</em> sensitive features. In other
-                words, features that differ significantly between groups act as proxies
-                for the sensitive feature.</pr>
-                </details>
-                """, 'markdown')
-        return cell
+            self.figs.append(self._create_chart(f, BALANCE_DESCRIPTION))
```
amrasekh, line 131: You meant `BALANCE_METRICS_DESCRIPTION`? `BALANCE_DESCRIPTION` does not seem to be defined anywhere.

credoai/artifacts.py
```diff
@@ -164,18 +164,18 @@ def export_assessment_results(self,
         assessment_results : dict or list
             dictionary of metrics or
             list of prepared_results from credo_assessments. See lens.export for example
+        reporter_assets : list, optionall
```
amrasekh, line 167: typo: `optionall`

credoai/utils/lens_utils.py
```diff
@@ -0,0 +1,12 @@
...
+    Connects prepared results from a module to metric keys generated by Lens
+
+    Parameters
+    ----------
+    module_prepared_results : DataFrame
+        output of CredoModule.prepare_results()
+    lens_prepared_results : DataFrame
+        output of lens._prepare_results(assessment)
+    """
+    return module_prepared_results.merge(lens_prepared_results)['metric_key'].tolist()
```
amrasekh, line 12: Isn't this missing `on` parameter? Column name to join on.

amrasekh (changes requested):
Some tests also need to be updated; some are failing.

credoai/utils/lens_utils.py
IanAtCredo: it's automatically "on" all overlapping columns. But I've moved away from this function anyway

amrasekh (approved):

PR #134, opened 2022-06-08
Security module

amrasekh:
This adds the security module covering the assessment for two types of attacks:
* Model extraction: already included
* Model evasion: being included in this PR as well


credoai/modules/model_modules/security.py
```diff
@@ -0,0 +1,145 @@
...
+        In model extraction, the adversary only has access to the prediction API of a target model
+            which she queries to extract information about the model internals and train a substitute model.
+
+        Returns
+        -------
+        dict
+            Key: extraction_attack_score
+            Value: accuracy of the thieved model / accuracy of the victim model, corrected for chance
+        """
+        # use half of the test data for model extraction and half for evaluation
```
IanAtCredo, line 96: I'm guessing extraction attack success is a function of the amount of data that is used?

Right now the extraction attack will assume that the attacker has access to half X_test. Why are we making that assumption? Why not use cross-validation?

credoai/modules/model_modules/security.py
```diff
@@ -0,0 +1,145 @@
...
+        The test features
+    y_test : pandas.Series
+        The test outcome labels
+    """
+
+    def __init__(
+        self, model, x_test, y_test
+    ):
+        self.x_test = x_test.to_numpy()
+        self.y_test = to_categorical(y_test, num_classes=2)
```
IanAtCredo, line 40: Couldn't most of this functionality work for multiclass classification? Why are we limiting it to binary classification rather than inferring the number of classes?

credoai/modules/model_modules/security.py
```diff
@@ -0,0 +1,145 @@
...
+        y_true = [np.argmax(y, axis=None, out=None) for y in y_test]
+
+        y_pred = [np.argmax(y, axis=None, out=None) for y in thieved_classifier._model.predict(x_test)]
+        thieved_classifier_acc = sk_metrics.accuracy_score(y_true, y_pred)
+
+        y_pred = self.victim_model._model.predict(x_test)
+        victim_classifier_acc = sk_metrics.accuracy_score(y_true, y_pred)
+
+        metrics = {
+            'extraction_attack_score': max((thieved_classifier_acc - 0.5) / (victim_classifier_acc - 0.5), 0)
```
IanAtCredo, line 126: Nit: Are you organizing this as a dictionary because you intend (or there is the potential) to add other scores in the future?

credoai/modules/model_modules/security.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Returns
+        -------
+        dict
+            Key: metric name
+            Value: metric value
+        """
+        extraction_performance = self._extraction_attack()
+
+        attack_scores = {
```
IanAtCredo, line 56: Nit: I think you can just say `self.results = {...}`, rather than setting all these unnecessary variables.

IanAtCredo (commented):
Added some questions

credoai/modules/model_modules/security.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # extract
+        copycat = CopycatCNN(
+            classifier=self.victim_model,
+            nb_epochs=5,
+            nb_stolen=len_steal
+            )
+
+        thieved_model = self._get_model(x_steal.shape[1])
+        thieved_classifier = KerasClassifier(thieved_model)
```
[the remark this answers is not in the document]
amrasekh: Yes. And ART leaves it to a user to design and pass.
It needs to be a tensorflow model. [Here](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/attacks/extraction/copycat_cnn.py#L119) is where it is trained in ART.

credoai/modules/model_modules/security.py
amrasekh: I also noticed this issue. A reasonable process would be one in which an attacker obtains highest accuracy with lowest inference queries through smart model design and querying.
But ART requires the feature vectors to be passed beforehand (I think the attack itself should identify the optimal feature vectors to query for, similar to active learning). All the [three ART extraction methods](https://adversarial-robustness-toolbox.readthedocs.io/en/latest/modules/attacks/extraction.html) require that.

credoai/modules/model_modules/security.py
amrasekh: Correct. ART for example already has two other extraction methods that can be added. And of course there are other performance metrics we can add.

credoai/modules/model_modules/security.py
amrasekh: Agreed. Done.

credoai/modules/model_modules/security.py
amrasekh: I feel positive that it could. I go through the code to see what needs to be changed and upgrade the module accordingly.

credoai/modules/model_modules/security.py
IanAtCredo: Makes sense - that means we need to make the choice for them. We are choosing to use half of X_test. We should list these parameterizations somewhere. I'll make a confluence doc.

credoai/modules/model_modules/security.py
IanAtCredo: Cool, let's make a follow-up ticket quickly to extend this module

credoai/modules/model_modules/security.py
```diff
@@ -0,0 +1,212 @@
...
+        return metrics
+
+    def _evasion_success_rate(self, l1, l2):
+        """Calculates evasion success rate
+
+        Parameters
+        ----------
+        l1 : list
+            predictions of the original samples
+        l2 : _type_
```
IanAtCredo, line 201:
```suggestion
        l2 : list
```

credoai/modules/model_modules/security.py
```diff
@@ -0,0 +1,212 @@
...
+            predictions of the adversarial samples
+
+        Returns
+        -------
+        float
+            the proportion of the predictions that have been flipped
+        """
+        if len(l1) > 0:
+            return np.count_nonzero(l1 != l2) / len(l1)
```
IanAtCredo, line 210: `l1 == l2` will only work for numpy arrays, but the function signature says lists. Could require numpy arrays or use `np.equal`

IanAtCredo (changes requested):
Few small bugs. Testing needs to work. I haven't run it yet, so will do that in a bit but wanted to get a review out with most of what I think needs to change.

credoai/modules/model_modules/security.py
```diff
@@ -0,0 +1,223 @@
...
+        adver_pred = self.victim_model._model.predict(adver_sample)
+
+        # standardize for robust distance calculation
+        scaler = StandardScaler()
+        scaler.fit(self.x_train)
+        origl_sample_scaled = scaler.transform(origl_sample)
+        adver_sample_scaled = scaler.transform(adver_sample)
+
+        # generate success rate vs median distance threshold curve
+        distances = np.diag(euclidean_distances(
```
IanAtCredo, line 188: I changed this distance calculation to use numpy. It's faster and less error prone. There was also a "median" function that wasn't doing anything from what I could see. You'd calculate the norm between two vectors (which is a scalar) and then take the median. If I'm missing something please correct!

IanAtCredo (approved):
Please just look at distance metric change

credoai/modules/model_modules/security.py
amrasekh: Ticket created in the backlog.

credoai/modules/model_modules/security.py
amrasekh: Good catch. I did intend the median being there, but it seems it was not doing what it is supposed too. The idea is for the distance to be insensitive to the feature vector's length. Perhaps we can normalize the current distance by the length of the vector.

PR #137, opened 2022-06-15
fixed concatenation issue leading to incorrect infographic aspects

IanAtCredo:
## Change description
This bug:
image

Was related to the sensitive feature dataframe not relating to the X. I've fixed it within the fairness module, though I think updating credodata would be better long term.


amrasekh:
The report plots have not been displayed in my jupyter notebooks recently. Adding this below fixed it. Can consider adding it to the quickstart:

`%matplotlib inline`

PR #143, opened 2022-06-29
changed how errors with assessment plan are handled

IanAtCredo:
## Change description
When assessment plans weren't set up correctly (e.g., no metrics defined for performance assessment), Lens would fail.

This replaces that behavior with a note about failed assessment initialization, while printing the error. Otherwise things move on.


amrasekh (changes requested):
This PR has caused the 'validation_training_model' bunch to be missing from the results for when it should be included. I am also looking into it.

`lens.get_results()["validation_training_model"]`
`KeyError: 'validation_training_model'`

amrasekh (approved):

PR #154, opened 2022-07-28
Feat/munichre register

IanAtCredo:
## Change description
This adds the ability to apply a template programatically. Also added information about that functionality to the governance integration notebook.


credoai/artifacts.py
```diff
@@ -110,36 +119,43 @@ def get_assessment_plan(self):
...
 
     def get_defined_ids(self):
         """Return IDS that have been defined"""
         return [k for k, v in self.get_info().items() if v]
 
-    def register(self,
-                 model_name=None,
-                 dataset_name=None,
-                 training_dataset_name=None):
+    def register(
```
amrasekh, line 152: My understanding is that it applies/assigned a template to a model, and not register the template. If it is right, it is rather confusing that `assessment_template` is here.

amrasekh (approved):
In the governance integration demo, can consider an update like this to make it even easier for a user to use:

```
use_case_id = 'x'  # your use case id can be found in the Governance App
model_id = 'x'  # your model id can be found in the Governance App
governance = f'use_cases/{use_case_id}/models/{model_id}/assessment_spec' 

lens = cl.Lens(model=credo_model,
               data=credo_data,
               governance=governance
               )

lens.run_assessments().export()
```

credoai/artifacts.py
IanAtCredo: You are right, but I didn't want to require another function call. Gonna leave it like this for now.

PR #157, opened 2022-08-02
Add CredoApiClient

aproxacs:
## Change description
CredoApiClient is responsible to interact with Credo API server.
CredoApi has interface functions by using CredoApiClient
Changed Governance to use CredoApi.
Also added tests to protect core functions of CredoApi and CredoApiClient


IanAtCredo (approved):
I've gone through, tested and corrected some bugs.

One test is failing here, though not locally - not sure why. Maybe something to do with finding the default config location? Once that test is fixed this looks good-to-go

PR #158, opened 2022-08-02
Upgraded privacy module

amrasekh:
Upgraded privacy module to support multi-class and be architecture-agnostic.


IanAtCredo (changes requested):
Assessments need to be updated to show this works now for multiclass classification

PR #159, opened 2022-08-03
Upgraded privacy and security modules

amrasekh:
Upgraded privacy and security modules to support multi-class and be architecture-agnostic.

Comments to address:
- Is there a benefit to the SklearnClassifier? Should we use that one when the model is a sklearn classifier?
- Assessments need to be updated to show this works now for multiclass classification


IanAtCredo (changes requested):
Two things:
1. Doesn't `_get_model` need to be updated to deal with the multiclass setting?
2. Our assessments requirements currently state that binary classification is the only thing that's covered. This moves to remove that barrier, right? So we should update the assessments language.

amrasekh:
1. Updated `_get_model` in security module need to be to deal with the multiclass setting.
2. Updates assessment codes and docs to reflect the new (relaxed) requirements.
3. For the attacks currently used, `SklearnClassifier` does not seem to have any advantages over `BlackBoxClassifier` as they seem to only use `predict` function of the classifier. But I look deeper into them (needs reviewing ART's source codes).

IanAtCredo (approved):

PR #162, opened 2022-08-10
Feat/new credo data

IanAtCredo:
## Change description
Changed how CredoData works. Made it simpler and set it up to work with a broader range of data types in the future.

Other small changes:
* Removed customization of module "runs". Every customization should be in the module's initialization. If we need run level customization in the future we should build it, but removing unnecessary complexity for now.
* Updated schematic


credoai/artifacts.py
```diff
@@ -474,164 +414,144 @@ class CredoData:
...
+        data = {"X": self.X, "y": self.y, "sensitive_features": self.sensitive_features}
+        return data
+
+    def get_X_type(self):
+        return self.X_type
+
+    def get_y_type(self):
+        return self.y_type
+
+    def _determine_X_type(self):
```
fabrizio-credo, line 464: It seems var types are unused, for both x and y:
`_determine_X_type` -> `self.X_type` -> `get_X_type` -> this is not used anywhere that I can see

Also `self.X_type` is unused  since `isinstance` is used consistently across the class

credoai/artifacts.py
```diff
@@ -462,7 +402,7 @@ def predict_proba(X):
 class CredoData:
     """Class wrapper around data-to-be-assessed
 
-    CredoData serves as an adapter between tabular datasets
+    CredoData serves as an adapter between datasets
```
fabrizio-credo, line 405: Can we add a couple of lines here or at the bottom of the Docstring, specifying what
the expected types for X/y/sensitive features are after the whole validation + processing is completed?

fabrizio-credo (approved):
The index assignment seems OK to me. Just a couple of minor comments

credoai/artifacts.py
IanAtCredo: Agreed. Getter methods are unneeded. `X_type` is not intended to just be `isinstance` in the future. More like "tabular"

credoai/artifacts.py
IanAtCredo: Sure

PR #163, opened 2022-08-10
Feat/privacy

fabrizio-credo:
## Restructure of the privacy module

This is an attempt to generalize the code to run most type of privacy attacks from ART.
It is based on the premise that we have mostly 2 type of models, the ones requiring
fitting and the ones that don't.

## Changes
- Created a dictionary of maintained methods:

```
self.SUPPORTED_PRIVACY_ATTACKS = {
            "MembershipInferenceBlackBox": {
                "attack": MembershipInferenceBlackBox,
                "type": "model_based",
            },
            "MembershipInferenceBlackBoxRuleBased": {
                "attack": MembershipInferenceBlackBoxRuleBased,
                "type": "rule_based",
            },
        }
```
The idea is to be able to add an attack here, and it should run automatically as part of the test suite.
There is a way to infer the `type`, but I think it's better to be more deliberate.
This framework could be used also to specify other customization for the attacks if need be.
- Created a general method to run ART privacy modules
- Created a couple of static methods to further modularize logic


IanAtCredo (approved):
Looks good. I made a few minor changes (mainly added `train_test_split`).

PR #164, opened 2022-08-17
Feat/privacy att inf

fabrizio-credo:
## Attribute inference attacks added

Added:
- AttributeInferenceBaseline (uses data to predict one of the features)
- AttributeInferenceBlackBox (uses data plus attacked model prediction to predict one of the features)

## Logic changes

1. Expanded list and logic of supported attacks in the class
2. Updated `_general_attack_method` to account for the extra logic
3. Added `attack_feature` to PrivacyModule `__init__`, this is the position of the attribute to infer
4. Attribute attacks are off by default, to activate them instantiate PrivacyModule with a `attack_feature not None` 

## Retro-change
Rows 196-212 -> I reverted to the old way of splitting dataset, removing sklearn train_test_split -> this doesn't work if a train/test pair with different lengths is passed.


credoai/modules/model_modules/privacy.py
```diff
@@ -131,54 +181,73 @@ def _general_attack_method(self, attack_details):
...
+                        ]
+                    ).reshape(-1, 1)
+                }
+            original = x_test_bln[:, self.attack_feature].copy()
+            inferred = attack.infer(
+                np.delete(x_test_bln, self.attack_feature, 1), **extra_arg
+            )
 
-        return self._assess_attack(train, test, accuracy_score)
+            return np.sum(inferred == original) / len(inferred)
```
IanAtCredo, line 250: What will happen if there are multiple features? I think this calculation is wrong - this should be divided by the number of elements, not just the length. E.g., `np.size(inferred)`

However, even that just reports the attribute inference across all items. We may want to report attribute inference success differentially for each value. This form of the test says "if I have none of these attributes, how well can the system infer them". Do you think that's the better test than "if I have all but one of these features, how well can I infer that feature"?

IanAtCredo (changes requested):
## Style

Most of my comments are stylistic. This became harder to read. Not necessary, but I think breaking up `_general_attack_method` into submethods could help.

`_general_attack_method` should definitely have keyword arguments that are documented, regardless.

I think you should move `self.SUPPORTED_PRIVACY_ATTACKS` into a separate constant. The components that are customized within the class can be referenced using the other class variables set when needed (`self.attack_score`, `self.attack_model`)

## Function
* The change back from `train_test_split` doesn't recreate the same functionality. SEe comment
* Before adding this I'd like to see some testing added. What scores are you expecting for attribute inference? Are you sure this is working? Are the scores sensible?
* You removed the "membership_inference_attack_score" it seems. This seems like chesterton's fence error. I.e., we didn't talk about exactly why this code was written as it was, but there was a reason!
     * Specific metric names are associated with specific risk issues via our [rai-service](https://github.com/credo-ai/rai-service/blob/develop/src/app/core/lens/metric_type_constants.py#L28). We felt like it was worth creating ONE membership inference attack score over all membership inference attacks.

fabrizio-credo:
I think I have addressed most of the points above. The major changes:
- Moved dictionary of methods outside class + streamlined the dictionaries
- Tidied up the run and general_attack methods
- Tested extensively comparing against ART code tutorial
- PrivacyModule is initialized with actual model, not `CredoModel` -> Modified the privacy assessment class to reflect this change

A point I would like to discuss:
- Behavior for multiple features attribute attack. I think it makes more sense to interpret this as a user wanting to test attacking different columns (corresponding to different binary features) in their dataset, rather than allowing for one hot encoded categorical. I can basically loop the attribute attack part of the code, and output a more complex result dictionary to display this information.

credoai/modules/model_modules/privacy.py
```diff
@@ -76,127 +124,195 @@ def run(self):
...
+
+        Returns
+        -------
+        tuple
+            Length 2 tuple, first elements contains the split of the train set,
+            the second element contains the split of the test set.
+        """
+
+        train_sets = train_test_split(
+            args[0], args[1], random_state=42, train_size=self.attack_train_ratio
```
IanAtCredo, line 256: You are passing a list of `args`, which should be generic, but then expecting them to be of some specific form. These are no longer general args. These are more like `training_arrays` and `testing_arrays`.

credoai/modules/model_modules/privacy.py
```diff
@@ -76,127 +124,195 @@ def run(self):
...
-        the rule based ones. The former require an extra fit step, and a further
-        split of tranining and test so that there is no leakage during the assessment
-        phase.
+        the rule based ones. The former require an extra fit step.
 
         Parameters
         ----------
         attack_details : dict
-            Map of all the supported ART privacy modules and their relative type.
+            Dictionary containing all the attack details
```
IanAtCredo, line 159: It's still not clear what `attack details` are. What are their components and how are they used? Rather than defining that, I think writing out the arguments here would be easier.

IanAtCredo (approved):
I made some non-blocking suggestions. If you think they are unnecessary, please merge and add a ticket to make these changes in the future.

credoai/modules/model_modules/privacy.py
IanAtCredo: It seems you want to keep them as "attack details" so you can pass them around. If so, that's fine, but you may want to create a namedtuple or dataclass

PR #169, opened 2022-09-07
changed to _setup scheme from __call__

IanAtCredo:
## Change description
See code - small changes, mostly changing `__call__` function to `_setup`


fabrizio-credo:
I made a change to privacy.py and evaluator to demonstrate a way of achieving validation with minor changes.

1. Evaluator contains a list of `required artifacts` (this is information needed by lens to pass only the needed arguments to the evaluator)
2. In evaluator.py: `self.__dict__.update(kwargs)` deals with the assignemnt of the artifacts
3. Validation must only deal with artifacts

PRO: Simple and minimal changes
CON: A developer must create the evaluator assuming that the artifacts will be available in self, there is no explicit assignemnt.

Alternative:
Split `_setup` in:
-> `_init_artifacts`: explicit initialization of the artifacts
-> `_setup`: does all the other operations
`_call` becomes:
1. _init_artifacts
2. _validate_arguments
3. _setup
PRO: everything is explicit CON: tedious assignment of the artifacts and an extra method that can be avoided

IanAtCredo:
I feel like the `__dict__` update works. You'll always have the `model`, `assessment_data`, `training_data` and whatever else in the future. `_setup` determines what is expected.

I'm going to define an `_init_artifacts` function, but it is not expected to be overwritten by child classes.

PR #171, opened 2022-09-13
changes

IanAtCredo:
## Change description
* Require that the results created by an evaluator is a list of containers. Cannot be a single container!
* Simplified Performance evaluator. There was functionality kept over from the previous module/assessment divide.
    * For instance, results were formatted by `prepare_results`. They can now be manipulated as dfs within the result generating functions. 
    * No need for `NotRunError`
 * Lens - simplified some things
     * Because all results will always be lists, could eliminate some if statements
     * `flatten_list` was already a util in common
     * Added a `get_evaluators` function

Evidence changes:
* The data property seems to obscure what the evidence is about. I added back the _struct function (which should be filled out by the children), and that should add keys to the struct dictionary. Passing a nested dictionary in data didn't seem correct.
* Added validation to make sure the table has a name
* Removed ID from evidences. Doesn't seem necessary

## ToDo
* Do similar simplification for other evaluators

### DIFF
--- a/credoai/evaluators/evaluator.py
+++ b/credoai/evaluators/evaluator.py
@@ -23,13 +23,11 @@ class Evaluator(ABC):
 
     @results.setter
     def results(self, results):
-        # TODO: validation to add after evaluators are refactored
-
-        # if not isinstance(results, list):
-        #     raise ValidationError("Results must be a list")
-        # for result in results:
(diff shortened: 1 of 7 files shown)


IanAtCredo:
@fabrizio-credo , if I am missing things with this PR, let's discuss.

PR #175, opened 2022-09-21
Feat/governance in lens

IanAtCredo:
Number of changes:
* Governance accessible within Lens
* required artifacts -> sets
* Evaluators have the ability to create "container_info" which will be passed to containers to inform their metadata/additional labels
* some updates to __init__.py files to make imports easier
* Added metadata to EvidenceContainers, to be passed similar to labels
* Changes some aspects of Table evidence
* Changes Profiler to output json instead of csv and added validation for form of profiler
* Lens
    * Reordered private methods
    * added `send_to_governance` method
    * Removed labels passing to evidence via Lens - should be done within the evaluator


credoai/governance/governance.py
```diff
@@ -140,7 +142,7 @@ def register(
             self.set_evidences([])
 
     def __parse_json_api(self, json_str):
-        return deserialize(json.loads(json_str))
+        return json.loads(json_str)
```
IanAtCredo, line 145: @aproxacs, I was getting an error with deserialization when using the file taken straight from the platform.

image

I think it's because the platform already deserializes before downloading. Is that true? If so, this fixes the bug.

aproxacs: yeah, right, I am fixing this. UI app should return original JSON API format.

IanAtCredo: Ok, I'll change this code back then

PR #176, opened 2022-09-21
feat(): export  assessment to file

aproxacs:
## Change description
`gov.export(filename)` saves assessment to the filename. 
The format is json api. So the UI app just need to pass this data to the API server. 
It also saves Lens version in meta. So that UI app and API server verifies that it is generated from Lens.

### DIFF
--- a/credoai/governance/governance.py
+++ b/credoai/governance/governance.py
@@ -3,12 +3,14 @@ Credo Governance
 """
 
 import json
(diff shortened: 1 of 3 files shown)


aproxacs:
UI updates PR : https://github.com/credo-ai/credo-ui/pull/586

IanAtCredo (approved):

PR #177, opened 2022-09-21
Ranking fairness evaluator

amrasekh:
Adds ranking fairness evaluator

It calculates three metrics for ranking use cases where the ranking and sensitive feature vectors are available.


fabrizio-credo (approved):
Feel free to merge!

PR #180, opened 2022-09-23
Feat/governance ingestion

IanAtCredo:
* Added ModelEquity evaluator
* renamed `shared` in utils -> `fairlearn`
* Changed how labels are created for Evidence. Also added @property to evidence subclasses. @fabrizio-credo , it seems even if you declare an abstract property in a baseclass, you need to re-declare a @property in the child. Otherwise the method is treated as a method, not a property. E.g., you can't write "foo.bar". You'd have to write "foo.bar()"
* Added the ability for governance to check evidence against requirements before export

## Pipeline Creator
* Added PipelineCreator, which can create a pipeline from governance
* Moved old "automatic" pipeline creation to pipeline creator.
* Removed and/or shuffled functions around lens to accomodate new pipeline creator

### DIFF
--- a/credoai/evaluators/utils/utils.py
+++ b/credoai/evaluators/utils/utils.py
@@ -1,0 +1,67 @@
+import inspect
+import sys
+
+
+def list_assessments_exhaustive():
+    """List all defined assessments"""
+    return inspect.getmembers(
+        sys.modules[__name__],
+        lambda member: inspect.isclass(member) and member.__module__ == __name__,
+    )
+
+
+def list_assessments():
+    """List subset of all defined assessments where the module is importable"""
+    assessments = list_assessments_exhaustive()
+    usable_assessments = []
+    for assessment in assessments:
+        try:
+            _ = assessment[1]()
+            usable_assessments.append(assessment)
+        except AttributeError:
+            pass
+    return usable_assessments
+
+
+def get_assessment_names():
+    return {name: assessment_class().name for name, assessment_class in ASSESSMENTS}
+
+
+def get_assessment_requirements():
+    return {
+        name: assessment_class().get_requirements()
+        for name, assessment_class in ASSESSMENTS
+    }
+
+
+def get_usable_assessments(
+    credo_model=None,
+    credo_data=None,
+    credo_training_data=None,
+    candidate_assessments=None,
+):
+    """Selects usable assessments based on model and data capabilities
+
+    Parameters
+    ----------
+    credo_model : CredoModel
+    credo_data : CredoData
+    candidate_assessments : list, optional
+        list of assessments to use. If None, search over all
+        possible assessments, by default None
+
+    Returns
+    -------
+    dict
+        dictionary of assessments, {name: assessment, ...}
+    """
+    if candidate_assessments is None:
+        to_check = [a[1] for a in ASSESSMENTS]
+    else:
+        to_check = candidate_assessments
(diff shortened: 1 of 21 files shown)


credoai/evaluators/utils/utils.py
```diff
@@ -0,0 +1,67 @@
...
+        try:
+            _ = assessment[1]()
+            usable_assessments.append(assessment)
+        except AttributeError:
+            pass
+    return usable_assessments
+
+
+def get_assessment_names():
+    return {name: assessment_class().name for name, assessment_class in ASSESSMENTS}
```
fabrizio-credo, line 27: `ASSESSMENTS` seems to be missing. In general, is the code here used anywhere atm, or is this more like a placeholder?

credoai/lens/lens.py
```diff
@@ -60,14 +60,13 @@ def __init__(
...
         self.pipeline_results: list = []
-        self._validate()
         if self.assessment_data and self.assessment_data.sensitive_features is not None:
             self.sens_feat_names = list(self.assessment_data.sensitive_features)
         else:
             self.sens_feat_names = []
-        # If a list of steps is passed create the pipeline
-        if pipeline:
-            self._generate_pipeline(pipeline)
+        self._generate_pipeline(pipeline)
```
fabrizio-credo, line 67: This is running everytime now, the code inside `_generate_pipeline()` does not account for an empty pipeline.
In `_generate_pipeline()` we could for example add something like:
```
if not pipeline:
  return
```

credoai/lens/lens.py
```diff
@@ -331,21 +331,21 @@ def _generate_pipeline(self, pipeline):
         ValueError
             Id needs to be a string.
         """
+        if pipeline is None and self.gov:
```
fabrizio-credo, line 334: The default for `pipeline` is an empty dictionary, `{} is None` would return False always I think.

credoai/lens/lens.py
```diff
@@ -376,13 +376,20 @@ def _validate(self):
                     raise ValidationError(
                         "Sensitive features should have the same shape across assessment and training data"
                     )
+        if len(self.pipeline) == 0 and self.gov is None:
```
fabrizio-credo, line 379: `_validate` is running after Lens init, in theory we can have an empty pipeline and no governance, this is when users add steps manually.

credoai/lens/lens.py
```diff
@@ -162,11 +161,6 @@ def run(self):
         """
         Run the main loop across all the pipeline steps.
         """
-        if len(self.pipeline) == 0:
```
fabrizio-credo, line 165: This is connected to the comment on `_validate()`, if we are still allowing the user to manually add steps 1 by 1, we can still get here and have empty pipeline, if that is the case the if statement can be use to collect the error, rather then generate the pipeline like it used to before.

fabrizio-credo (changes requested):
Everything is looking good, minor comments:

- I think a couple of changes need to be made in Lens validation process, to handle the user adding steps one by one to the pipeline, assuming this is a behaviour we want to retain.

credoai/lens/lens.py
IanAtCredo: I removed this validation

credoai/lens/lens.py
IanAtCredo: Yep, removed.

Issue #183, opened 2022-09-30
Probability-dependent metrics do not run (not added to pipeline)


esherman-credo:
## Expected Behavior
Probability-dependent metrics, like `roc_auc_score` and `average_precision_score`, should produce results similar to other scalar metrics (e.g. `accuracy_score`).

## Actual Behavior
Adding probability-dependent metrics to pipeline throws an error (see below) with stack trace originating in this file:
https://github.com/credo-ai/credoai_lens/blob/release/1.0.0/credoai/evaluators/utils/fairlearn.py

<img width="746" alt="image" src="https://user-images.githubusercontent.com/114082449/193288183-ed49ac05-76fc-4ccc-b2e4-c6245745cd50.png">

## Steps to Reproduce the Problem

  1. Add `'roc_auc_score'` to evaluator pipeline in `Quickstart.ipynb`. See `metrics =` line in below screenshot. This cell is the first code cell below "Lens in 5 minutes" header.

<img width="745" alt="image" src="https://user-images.githubusercontent.com/114082449/193288358-f7ddc3e5-7e13-43ed-b476-980edb754b62.png">

  2. Run all cells in notebook up to and including pipeline cell pictured above.

## Specifications
release/1.0.0 branch

PR #184, opened 2022-09-30
Fix/183 probability dependent metrics do not run not added to pipeline

esherman-credo:
## Change description
1) Add `y_true` to `create_metric_frames()` call in `fairlearn.py` within if-statement for probabilistic metrics
2) Unindented if-block (see line 30 in fix branch), since probabilistic metrics are not necessarily performance metrics (putting `roc_auc_score` in Quickstart.ipynb pipeline alone without other metrics will make it so `roc_auc_score` doesn't get calculated).
3) Added check for whether the specified metric is meant to be for a binary prediction problem (rather than multi-output classification/probabilistic output). Sklearn's `predict_proba()` function outputs an array that has shape `(n_samples, n_classes)`. For binary problems, `y_true` will be `(n_samples, )` in our present handling, and `(n_samples, n_classes)` for a multi-output problem. The former causes a shape mismatch when calling the scoring function and so we need to select the column of `y_prob` corresponding to the positive class.
See discussion of `y_score` here: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html

## Type of change
- [X ] Bug fix (fixes an issue)

## Related issues

> Fix https://github.com/credo-ai/credoai_lens/issues/183

### DIFF
--- a/credoai/evaluators/utils/fairlearn.py
+++ b/credoai/evaluators/utils/fairlearn.py
@@ -26,12 +26,12 @@ def setup_metric_frames(
             sensitive_features=sensitive_features,
         )
 
-        # for metrics that require the probabilities
-        prob_metric_frame = None
-        if y_prob is not None and prob_metrics:
-            metric_frames["prob"] = create_metric_frame(
-                prob_metrics,
-                y_prob,
-                sensitive_features=sensitive_features,
-            )
+    if y_prob is not None and prob_metrics:
+        metric_frames["prob"] = create_metric_frame(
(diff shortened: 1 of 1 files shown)


IanAtCredo:
I don't think we should handle the transformation of `predict_proba` at metric_frame time. We should expect that the predictions created by the model work for the metric. If they don't, I feel the prediction function should be changed. 

Let's table that for more discussion and make this PR just about the bug fix. I'll update.

PR #185, opened 2022-09-30
update predict_proba when binary classification in sklearn framework

IanAtCredo:
## Change description

Trying to deal with issue with predict_proba leading to possible issues in binary case.


esherman-credo (approved):
lgtm

PR #187, opened 2022-10-03
Feat/len faq

IanAtCredo:
## Change description
Documentation updates


fabrizio-credo (approved):
Looks good to me. 

Just to be pedantic, when I tested the new requirements installation I used `pip`.
I haven't tried with `conda` since I don't normally use it, but I am assuming that everything should work the same.

PR #191, opened 2022-10-05
fixed processing and validation that were somehow lost

IanAtCredo:
## Change description
For some reason validation and processing was removed during some previous PR. adding back in.

### DIFF
--- a/credoai/artifacts/data/base_data.py
+++ b/credoai/artifacts/data/base_data.py
@@ -47,11 +47,11 @@ class Data(ABC):
         sensitive_intersections: Union[bool, list] = False,
     ):
         self.name = name
-        self.X = self._process_X(X)
-        self.y = self._process_y(y)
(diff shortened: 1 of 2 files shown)


fabrizio-credo (approved):
@IanAtCredo the latest changes broke the test for ranking fairness, the reason is that in the example Amin created the assessment artifact is missing `X`. Ranks are passed as `y` directly.

I think by design we intended X, y and sensitive features to all be optional in the creation of **Data** artifact, so I fixed the issues according to that idea.

IanAtCredo:
> @IanAtCredo the latest changes broke the test for ranking fairness, the reason is that in the example Amin created the assessment artifact is missing `X`. Ranks are passed as `y` directly.
> 
> I think by design we intended X, y and sensitive features to all be optional in the creation of **Data** artifact, so I fixed the issues according to that idea.

I see. I actually thought X would always be passed, but this seems fine!

PR #193, opened 2022-10-06
Refact/remove prepare results

fabrizio-credo:
## Change description

- Removed `_prepare_results` from all evaluators
- Moved the error handling for missing results to the `results` property setter in **evaluator.py**
- Removed all referencing to `self._results` from evaluators because it bypasses the setter method


credoai/evaluators/evaluator.py
```diff
@@ -18,7 +18,12 @@ def __init__(self):
 
     @property
     def results(self):
-        return self._results
+        if self._results:
+            return self._results
+        else:
```
IanAtCredo, line 23: Nice - love the addition of more error messages

IanAtCredo (approved):

PR #195, opened 2022-10-06
Feat/table types

IanAtCredo:
## Change description

Change the "columns "part of TabelEvidence to return types for the columns. This functionality is there to support the Platform to style the column outputs (e.g., rounding numbers)

Example:
```
'columns': [
  {'value': 'false_positive_rate', 'type': 'number'},
   {'value': 'true_positive_rate', 'type': 'number'},
   {'value': 'threshold', 'type': 'number'}
]
```

### DIFF
--- a/credoai/evidence/evidence.py
+++ b/credoai/evidence/evidence.py
@@ -138,8 +138,11 @@ class TableEvidence(Evidence):
 
     @property
     def data(self):
+        columns = [
+            {"value": k, "type": str(v)} for k, v in self._data.dtypes.iteritems()
+        ]
         return {
-            "columns": self._data.columns.tolist(),
+            "columns": columns,
             "value": self._data.values.tolist(),
         }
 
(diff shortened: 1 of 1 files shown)


credoai/evidence/evidence.py
```diff
@@ -148,6 +151,17 @@ def base_label(self):
         label = {"table_name": self.name}
         return label
 
+    def _transform_type(self, pandas_type):
+        lookup = {
+            'int64': 'number',
+            'float64': 'number',
+            'object': 'str',
```
aproxacs, line 158: Can we use 'string' instead of 'str'

fabrizio-credo:
@IanAtCredo @aproxacs Is this pull request still meaningful?

aproxacs:
@fabrizio-credo Yes, please merge this. Now Backend converts this for old Lens versions, but it will be deprecated by the end of this year. 

Note : please change `str` to `string`

PR #201, opened 2022-10-13
FEAT/shap_evaluator

fabrizio-credo:
## Version 1 of the shap evaluator completed.

Functionality:
1. Output of summary statistics, currently mean and mean(|x|)
2. Optional output of shap values for specific samples in the dataset


credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+                TableContainer(ind_res, **self.get_container_info(labels=labels))
+            ]
+        return self
+
+    def _setup_shap(self):
+        """
+        Setup the explainer given the model and the feature dataset
+        """
+        explainer = Explainer(self.model.predict, self.X)
+        self.shap_values = explainer(self.X)
```
esherman-credo, line 62: Maybe add a comment clarifying that this is now expressly linked to an object of the (external) SHAP library.

credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        Return shapley values for multiple samples from the dataset.
+
+        Returns
+        -------
+        DataFrame
+            Columns:
+                values -> shap values
+                ref_value -> Reference value for the shap values
+                    (generally the same across the dataset)
```
esherman-credo, line 102: What exactly is this? The mean shap value for the given feature or expected value for a sample (i.e. mean prediction output for all samples)? 

Can we clarify the comment by pulling from SHAP's documentation?

credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+            self.results += [
+                TableContainer(ind_res, **self.get_container_info(labels=labels))
+            ]
+        return self
+
+    def _setup_shap(self):
+        """
+        Setup the explainer given the model and the feature dataset
+        """
+        explainer = Explainer(self.model.predict, self.X)
```
esherman-credo, line 61: It seems that you're just using SHAP's model-agnostic explainer (https://shap.readthedocs.io/en/latest/tabular_examples.html#model-agnostic). If I'm not mistaken the library has different SHAP algorithms for different base models so there might be some computational benefit to actually specifying Tree vs. linear model vs. something else.

Maybe something to track via Lens's model wrappers so that we can use here to better select the SHAP explainer?

credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+                values -> shap values
+                ref_value -> Reference value for the shap values
+                    (generally the same across the dataset)
+                sample_pos -> Position of the sample in the dataset
+        """
+        all_sample_shaps = []
+        for ind in self.samples_ind:
+            all_sample_shaps.append(
+                {
+                    **self._get_single_sample_values(self.shap_values[ind]),
```
esherman-credo, line 109: How is this different from `Explainer.explain_row()` ? 

See https://shap.readthedocs.io/en/latest/generated/shap.Explainer.html#shap.Explainer.explain_row

credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+        stats_1 = values_df.apply(["mean"]).T
+        stats_2 = values_df.apply(lambda x: mean(abs(x)))
+        stats_2.name = "mean(|x|)"
+        final = concat([stats_1, stats_2], axis=1)
+        final = final.sort_values("mean(|x|)", ascending=False)
+        final.name = "Summary of Shap statistics"
+
+        return final
+
+    def _get_mult_sample_shapley_values(self) -> DataFrame:
```
esherman-credo, line 92: One major concern about this is that it could reveal too much information about the underlying dataset nor the internals of the model.
For instance, we don't want someone to generate a report that has a full sample-by-sample output of how the model made it's predictions. It'd be a very long report so I think it's unlikely that someone would do that. I also don't know exactly how we can arbitrate what people do/don't try (is 5 individual samples too many? 10?).

Something to think about...

Another thought is whether it makes sense to intersect this (and how to do it) with groupby functionalities we've implemented elsewhere. For instance, maybe it makes more sense to allow a report to be generated with SHAP values based on males/females separately than it does to worry about individual-sample-SHAP

^ could be either a) mean/abs_mean shap values calculated on an all-male or all-female dataset or b) calculating shap values on a combined dataset and then reporting means from subsets (not sure off-hand if those will be mathematically guaranteed to be the same or different...)

credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(self, samples_ind: Optional[List[int]] = None):
+
+        super().__init__()
+        self.samples_ind = samples_ind
+
+    name = "Shap"
+    required_artifacts = ["assessment_data", "model"]
+
+    def _setup(self):
+        self.X = self.assessment_data.X
```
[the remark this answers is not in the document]
IanAtCredo: `data` is interpreted by Lens to mean it can work on training or assessment data. Fabrizio was intending this to only work on testing data

credoai/evaluators/shap.py
IanAtCredo: Agreed. If we can do smart inference based on what's in the `modelinfo` we should use it. Right now `modelinfo` only infers framework.

We can also add an initialization parameter that takes a SHAP explainer. We default to the less-than-optimal model-agnostic explainer otherwise.

IanAtCredo, line 61: I have also seen that the data passed is to estimate a "background distribution". I've sometimes seen a sample passed (more computationally efficient?) and sometimes seen training data passed. Not sure what is actually supposed to be done here.

credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def _get_overall_shap_contributions(self) -> DataFrame:
+        """
+        Calculate overall SHAP contributions for a dataset.
+
+        The output of SHAP package provides Shapley values for each sample in a
+        dataset. To summarise the contribution of each feature in a dataset, the
+        samples contributions need to be aggregated.
+
+        For each of the features, this method provides: mean, max, minimum and the
```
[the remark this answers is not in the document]
IanAtCredo: We should remove this from the comment. We had discussed max/min before (I'd care more about 1%/99% or 5%/95% - max/min is too outlier-y), but it was removed.

IanAtCredo (changes requested):
@esherman-credo had some good comments. Please follow up on his documentation requests. Also curious what your perspective is on the `Explainer` comments Eli brought up.

IanAtCredo:
I was using this today and it takes a looong time. That seems to be a common complain for Shap. This lead to me reading and I have a few comments/recommendations.

## Background Data
The long computation time is, amongst other things, dependent on algorithm type and the size of the background data. The background data is what is passed to the explainer and does _not_ have to be the whole dataset.

Here's a link to one of [multiple discussions](https://github.com/slundberg/shap/issues/1018). The two utils are `shape.sample` and `shap.kmeans`. Kmeans seems good (with maybe 50 centroids?) on first blush.

## Categorical Data
Digging deeper, there is a discussion about handling [categorical data](https://github.com/slundberg/shap/issues/451). Not sure what the takeaway here is. Not sure yet how kmeans is handling categorical. We could use kmeans if there are no categorical, and sample 100 points when there is categorical data. I haven't implemented this yet.

## Model-agnostic explainer
The explainer will try to automatically discover the model, and use a good algorithm, but only if you pass the model. If you pass the "predict" function, it will never make use of better models. We could try and use `model-like` and fall back if it fails.


## Conclusion

I've made changes based on what I said above, but I'm not sure if they are the best. Would love to hear your thoughts.

credoai/evaluators/shap.py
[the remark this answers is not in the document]
esherman-credo: That's fair. I was more just getting at `X` doesn't make clear that it's assessment vs. train vs. something else.

esherman-credo:
What were the sample sizes you were working with that ran slowly, Ian?
I guess for the sake of discussion let's pretend everything is using KernelSHAP. I think  many of these issues go away if we're using an algorithm-specific SHAP (like Tree; categorical becomes a non-issue there).

It feels a little odd to have kMeans be our default for under-the-hood speedup since it's hard to know whether your selection for k is "right". That applies to sampling too since sampling proportion is also a hyperparameter. Somehow sampling feels more natural, though? Maybe that's a cognitive bias of mine...I keep using the word "feels", after all.
Either way we should make it VERY clear in documentation what is going on. I could see a user being alarmed/confused that the SHAP beeswarm plot only has 50 points if their assessment data had 10M.

Separately, I also don't really understand what SHAP is doing when it encounters categorical or one-hot data, regardless of which algorithm it uses. It's probably just treating them as continuous and going on its merry way, which isn't great. The discussion in the 2nd link you gave didn't shed too much light on this.
Anecdotally, I ran into this issue in a healthcare paper I wrote a while back. We had some tertiary variables (like discharged to home vs. step-down unit vs. nursing home) which we one-hotted for the purposes of fitting a good model. The problem was that resulted in having 3 binary variables in the SHAP plot we generated. Listing them separately is a bit misleading because they could individually each be ranked higher than some other important variables but really it's all one conceptual variable. Plus there's mutual exclusion in how the samples are plotted in the beeswarm. For that project we ended up having to provide a somewhat stilted explanation of how to interpret the plot. Idk if it possible for a DS or Governance lead to put in that context next to their plot in a platform-generated report.

How safe of an assumption is it that our DS users will be always one-hotting their categorical variables? That seems like an obvious always-do to me.

credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class ShapValues(Evaluator):
+    """
+    This evaluator perform the calculation of shapley values for a dataset/model,
+    leveraging the SHAP package.
+
+    It calculates 2 types of assessments:
+    1. Overall statistics of the shap values across all samples: mean and mean(|x|)
+    2. Optional: individual shapley values for a list of samples
```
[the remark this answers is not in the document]
fabrizio-credo: Overall are always calculated at a dataset level.
The sample shap values are only calculated when a list of samples of interest is provided.

I will modify the doc though, it's not super readable.

IanAtCredo:
> the-hood speedup since it's hard to know whether your selection for k is "right". That applies to sampling too since sampling proportion is also a hyperparameter. Somehow sampling feels more natural, though? Maybe that's a cognitive bias of mine...I keep using the word "feels", after all. Either way we should make it VERY clear in documentation what is going on. I could see a user being alarmed/confused that the SHAP beeswarm plot only has 50 points if their assessment data had 10M.

## Background sampling

Kmeans is _not_ being used for creating SHAP values. It is being used to create a "background dataset" which is passed to the Explainer. I don't fully understand what that background is being used for - I believe it is basically used to impute values for a feature that is being masked for the purpose of SHAP estimation. The creator explicitly suggests using a small background dataset. The time SHAP takes to run is linear with the size of the background dataset, and there is some evidence that SHAP values stabilize at very low sample levels.

There isn't good guidance on kmeans vs sampling.

## Categorical data

Ideally, SHAP would use the encoding method the initial model used. But it doesn't seemto have access to that encoding. I don't think we can assume one-hot encoding. Creating categorical embeddings is a useful thing to do for neural networks, for instance.

credoai/evaluators/shap.py
```diff
@@ -0,0 +1,141 @@
...
+    def _get_overall_shap_contributions(self) -> DataFrame:
+        """
+        Calculate overall SHAP contributions for a dataset.
+
+        The output of SHAP package provides Shapley values for each sample in a
+        dataset. To summarise the contribution of each feature in a dataset, the
+        samples contributions need to be aggregated.
+
+        For each of the features, this method provides: mean, max, minimum and the
+        mean of the absolute value of the samples shapley values.
```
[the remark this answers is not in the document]
fabrizio-credo: Maybe samples', I meant it as plural :)

esherman-credo:
P.S. KMedians might be better? Who knows.

credoai/evaluators/shap.py
fabrizio-credo: This are the actual shap values. As in for 2 samples and 3 features

sample1 -> shap_val_1, ..., shap_val_3
sample2 -> shap_val_1, ..., shap_val_3

fabrizio-credo:
## Categorical Variables

> Based on digging (including looking at the original paper), I think the idea is that the user-supplied background data enables the algorithm to know what the "default"/"reference" value for each feature is. Importances and Shapley values are calculated by comparing the predictions when features take on their observed values vs. the reference values from the background data. Does that conform with your understanding?

Yeah that is my understanding as well. Based on the conversation @IanAtCredo mentioned above, the main issues seems to be precisely the calculation of the reference value when the categorical value is one hot (or ordinal) encoded. I think the standard approach SHAP uses is not suitable for all cases, hence the suggestions of some of the contributors. 

I don't think there is a final best approach, in [this reply](https://github.com/slundberg/shap/issues/451#issuecomment-607804767) it is suggested that the original dataset (pre-encoding) can be passed to SHAP as long as the model contains the data preparation stage basically. Which is something that we support in Lens, since `model-like` can be a sklearn pipeline for example. Although I am not entirely convinced that this approach will solve all problems, and I am unclear if SHAP can handle unencoded categorical values natively.

I think we should add a general warning to the interpretation of categorical variables within the evaluator, I don't think we should make official recommendations since it seems there is no consensus on the best approach.

## Sampling
Introducing subsampling of the initial dataset makes sense to speed up computation. I tend to agree with @esherman-credo about how **kmeans** "feels" like some extra done to the data, but in general as long as we document the reasoning behind it I am fine with it. Overall though I would make the choice of subsampling magnitude (both for the amount of samples or number of clusters in kmeans) available to the user. Enforcing 100 samples or 50 clusters seems to strong of a stance to be defendable in every scenario. 

## Proposed changes
Here's my proposed changes:
1. Address/document categorical variables SHAP interpretation by referencing the "official" thread. No explicit stance on this.
2. Allow for the user to define amount of subsampling if they wish so.  Provide default values: I have no particular opinion on the actual defaults. As far as the subsampling method, maybe we could link the default to the size of the Dataset, e.g., 1% of the number of samples. 

@esherman-credo @IanAtCredo Thoughts?

credoai/evaluators/shap.py
esherman-credo: I was referring to the reference value. That's something related to the background data set?

credoai/evaluators/shap.py
```diff
@@ -14,15 +16,16 @@ class ShapExplainer(Evaluator):
...
 
     Parameters
     ----------
     samples_ind : Optional[List[int]], optional
         List of row numbers representing the samples for which to extract individual
-        shapley values , by default None
+        shapley values. This must be a list of integer indices. The underlying SHAP
+        library does not support non-integer indexing
     background_kmeans : bool, optional
         If True, use SHAP kmeans to create a data summary to serve as background data for the
```
esherman-credo, line 30: Add a comment saying we are doing this because SHAP is slow (maybe even link to Scott Lundberg's advice on this from Github posts we linked in our discussion).

credoai/evaluators/shap.py
```diff
@@ -131,6 +136,25 @@ def _get_mult_sample_shapley_values(self) -> DataFrame:
...
+        limit : int, optional
+            Max number of samples allowed, by default 5
+
+        Raises
+        ------
+        ValidationError
+        """
+        if self.samples_ind is not None:
+            if len(self.samples_ind) > limit:
+                message = "The maximum amount of individual samples_ind allowed is 5."
```
esherman-credo, line 155: Maybe would be good to explain that we're doing this to avoid leaking too much PII. Theoretically someone could just iteratively call the sample function with different subsets of indices; not much we can do about that. It's at least good to highlight the "why?" so it doesn't seem arbitrary.

esherman-credo (changes requested):

credoai/evaluators/shap.py
esherman-credo: It doesn't look like there are test cases for SHAP. I'm curious what that might look like?

fabrizio-credo: Probably a general run test, parametrized on sampling strategy.

credoai/evaluators/shap.py
fabrizio-credo: Yeah as per our discussion yesterday, I can't find any specific reference in the docs as to how it's calculated.

IanAtCredo (approved):
lg

PR #206, opened 2022-10-18
feat(): add warning message for not linked evidences (AAA-1456)

aproxacs:
## Change description
assessments upload API now returns the result about if evidence is linked or not. [This BE PR](https://github.com/credo-ai/credo-backend/pull/840) is about it.

It shows warning messages when API returns not linked evidences. Used warning messages that Kyle suggested on [this ticket](https://github.com/credo-ai/credo-backend/pull/840)

### DIFF
--- a/credoai/governance/governance.py
+++ b/credoai/governance/governance.py
@@ -227,6 +227,7 @@ class Governance:
 
         if filename is None:
             self._api_export(evidences)
+            
         else:
             self._file_export(evidences, filename)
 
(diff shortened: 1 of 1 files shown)


IanAtCredo:
I'm confused about this.

There is almost always going to be non-linked evidence (Lens creates more evidence then is strictly needed by requirements). Why do we want to warn users about it? If we want to reduce evidence, we need to explicitly remove any evidence that isn't associated with a requirement. If we want that, we should just do that. Not warn the user.

PR #207, opened 2022-10-18
Bugfix/json

IanAtCredo:
## Change description
Quick fix to get profiler to jsonify.

Built off of shape. We'll merge wherever we need.


fabrizio-credo (approved):
@IanAtCredo the change is fine, but why are you merging into shap_evaluator rather than develop?

IanAtCredo:
It was laziness due to working off of shap evaluator. Let me go back and change

PR #209, opened 2022-10-20
Model zoo test freeze

esherman-credo:
## Change description
Added frozen model, data, assessment plan, and results for a binary classifier.

Added test which runs lens on the frozen data with the frozen model using the assessment plan. Compares the results on the current run to the frozen results.

Wee had discussed the possibility of developing frozen results for other model types, such as regression (see Model Zoo section of https://credo-ai.atlassian.net/wiki/spaces/MVP/pages/428933133/Product+Demo+Resources). At this time, we don't believe that is necessary. The goal is to reveal broad-strokes issues when we make changes to functions relating to Lens's pipeline handling, evaluator running, or results generating. The goal is NOT to evaluate ML models themselves; Regression would likely reveal these results just as well as binary classification.

In the longer term, we might consider adding frozen results for other evaluators (e.g. Security).

## Type of change


IanAtCredo (changes requested):
We had discussed a few things that I don't see in this PR - want to hear your thinking.

* We had discussed including a `test_results` for every separate evaluator. That seemed a helpful way to organize things
    * I'm guessing you chose not to do that because of the need to use the frozen assessment, models, etc. So you didn't want to include those frozen results in that form
* But one thing we wanted was for these tests to separately test the results for each evaluator. That doesn't seem possible right now. It seems once one evaluator fails the assert statement the whole `test_results` will fail. So we won't have visibility in all of the separate evaluators.
    * So my main request is make this happen. You may need to use the paramterize decorator to run `test_results` multiple times with different evaluators.

Other thoughts:

* Nit: This seems sufficiently different from the rest of `test_lens.py` to move into its own kind of file.
* STRING2EVALUATOR seems like a useful utility function. Why not add it as a general utility? Could put it in `evaluator/utils/utils.py`. Rather than use a dictionary, you can construct it using each evaluators "name" attribute. 
* Why just two evaluators? Why not run the whole pipeline?
    *  One reason I can imagine is some evaluators have some randomness that we haven't accounted for and we won't get consistent results. Is this what you found?

fabrizio-credo (changes requested):
Let's move to this structure for METRICS related tests. 

I agree with @IanAtCredo , this can live in its own file.

```
@pytest.mark.parametrize("metrics", TEST_METRICS, ids=TEST_METRICS_IDS)
def test_performance(
    credit_classification_model, credit_assessment_data, credit_training_data, metrics
):
    lens = Lens(
        model=credit_classification_model,
        assessment_data=credit_assessment_data,
        training_data=credit_training_data,
    )
    evaluator = Performance(metrics)
    lens.add(evaluator, "dummy")
    lens.run()
    assert lens.get_results()
```

IanAtCredo (approved):
Seems good. Fix conflicts and merge!

fabrizio-credo (approved):

PR #211, opened 2022-10-21
Feat/feature drift

fabrizio-credo:
## Feature Drift

Implemented population stability index calculation for model predictions and across dataset features (optional)


credoai/evaluators/feature_drift.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+    Parameters
+    ----------
+    buckets : int, optional
+        Number of buckets to consider to bin the predictions, by default 10
+    buckettype :  Literal["bins", "quantiles"]
+        Type of strategy for creating buckets, bins splits into even splits,
+        quantiles splits into quantiles buckets, by default "bins"
+    csi_calculation : bool, optional
```
esherman-credo, line 39: Shouldn't this be `psi_calculation`?

esherman-credo: Ah, I see. This is "characteristic stability index". Please clarify in the documentation that there are two different metrics

credoai/modules/credoai_metrics.py
```diff
@@ -341,3 +344,90 @@ def gini_coefficient_discriminatory(y_true, y_prob):
...
+
+    The common interpretations of the PSI result are:
+
+    PSI < 0.1: no significant population change
+    PSI < 0.25: moderate population change
+    PSI >= 0.25: significant population change
+
+    The number of buckets chosen and the bucketing logic affect the final result.
+
+    Based on the code in github.com/mwburke by Matthew Burke.
```
esherman-credo, line 373: Please also add a citation for the original paper, if possible.

esherman-credo (changes requested):
Code in commit looks good. Please address comment documentation.

Please also add test case(s). Can be deterministic, I think. Maybe 1 to test underlying metric from credoai_metrics file and 1 to test evaluator.

esherman-credo (approved):

PR #214, opened 2022-10-25
made 'name' attribute implicitly defined, added PipelineStep class, r…

### LINKED ISSUE #212: Specifying multiple data sets when creating lens object causes each evaluator to only be run once
## Expected Behavior
Lens should return results for each specified dataset, and for each evaluator (e.g. 2 data sets + 1 evaluator --> 2 sets of results).

## Actual Behavior
Lens is currently only returning one result set when both `assessment_data` and `training_data` are passed to the `Lens()` constructor. It appears that only the results from the `training_data` are processed (likely because that parameter is listed 2nd?)

Creating data objects as follows:
```
credo_model = ClassificationModel(
    'credit_default_classifier',
    model
)
train_data = TabularData(
    name = "UCI-credit-default-train",
    X=X_train,
    y=y_train,
    sensitive_features=sensitive_features_train
)
test_data = TabularData(
    name='UCI-credit-default-test',
    X=X_test,
    y=y_test,
    sensitive_features=sensitive_features_test
)
```

Creating pipeline and `Lens` object as follows
```
# pipeline scan be specifed using a sklearn-like style
metrics = ['accuracy_score', 'roc_auc_score']
pipeline = [
    (Performance(metrics), 'Performance Assessment'),
    (ModelFairness(metrics), "Fairness Assessment"),
]


# and added to lens in one fell swoop
lens = Lens(
    model=credo_model,
    assessment_data=test_data,
    training_data=train_data,
    pipeline=pipeline
)

lens.run()
```

Expected Results:
dictionary with 2 sets each of Performance and ModelFairness results. Running as above yields results that look like this (not enough entries and entries are clearly on training_data since performance is so strong):
image

Commenting out the line that passes `train_data` to `Lens()` constructor yields performance that looks like this:
image

IanAtCredo:
…efactored Lens to make use of pipeline step, fixed bug where metadata was dropped when multiple datasets or sensitive features were used, removed IDs for pipelines (fixing ID collision), added to pipeline get method

Fixes #212 

## Change description
Everything in commit message! Repeating here:
* made 'name' attribute implicitly defined
* added PipelineStep class, refactored Lens to make use of pipeline step
* fixed bug where metadata was dropped when multiple datasets or sensitive features were used
* removed IDs for pipelines (fixing ID collision)
* added to pipeline get method


credoai/lens/lens.py
```diff
@@ -39,9 +55,8 @@ def __init__(
             Training data, extra dataset used by some of the evaluators, by default None
         pipeline : Pipeline_type, optional, default None
             User can add a pipeline using a list of steps. Steps can be in 2 formats:
-            - tuple: max length = 3. First element is the instantiated evaluator,
-            second element is the step id (optional), third element (optional) is metadata (dict)
-            associated to the step.
+            - tuple: max length = . First element is the instantiated evaluator,
```
esherman-credo, line 58: max length = ?

esherman-credo (approved):
Looks good. See minor comment about comment for Lens constructor.

credoai/lens/lens.py
```diff
@@ -16,6 +17,21 @@
 ## TODO: Decide Metadata policy, connected to governance and evidence creation!
 
 
+@dataclass
+class PipelineStep:
+    evaluator: Evaluator
+    metadata: dict = None
```
fabrizio-credo, line 23: Typing hint should be `metadata: Optional[dict] = None`

credoai/lens/lens.py
```diff
@@ -195,37 +194,83 @@ def send_to_governance(self, overwrite_governance=False):
...
+        Returns
+        -------
+        List of PipelineSteps
+        """
+        to_check = metadata or {}
+        if evaluator_name:
+            to_check["evaluator"] = evaluator_name
+        return [p for p in self.pipeline if p.check_match(to_check)]
+
+    def get_results(self, evaluator_name=None, metadata=None) -> Dict:
```
fabrizio-credo, line 246: Return type should be `List[Dict]`

fabrizio-credo (approved):
Minor corrections, looks great in general, I like the PipelineStep class.
Overall the flow of the code is improved as well.

You have a knack for recursive functions 😄

credoai/lens/lens.py
```diff
@@ -281,30 +313,30 @@ def _add(
...
-                available_datasets = [
-                    n for n, a in vars(self).items() if "data" in n if a
-                ]
-                for dataset in available_datasets:
-                    labels = {"sensitive_feature": feat} if check_sens_feat else {}
-                    labels["dataset"] = dataset
-                    evaluator_arguments["data"] = vars(self)[dataset]
+                for dataset_label, dataset in self.get_datasets().items():
+                    additional_meta["dataset"] = dataset_label
+                    step = deepcopy(pipeline_step)
```
esherman-credo, line 329: One thing that occurs to me now that I'm a little bit deeper on the deepchecks evaluator is that this strategy of copying evaluators per-dataset is not always going to give us what we want.

If I'm not mistaken, all the evaluators we have right now either require 'data' (i.e. it's a single-dataset assessment) or BOTH 'assessment' and 'train' data to be specified. In the latter case, things are simple: if you don't provide both, the validation check fails. In the former, however, right now we just run the evaluator on one or both data sets. This line achieves that by copying the pipeline step and specifying that we should use the 2nd dataset if one is present.

Integrating deepchecks into Lens throws a bit of a wrench in this. Some checks in the package only require a single dataset while others require both train and test. Following the existing paradigm of specifying a `required_artifacts` dictionary, we could require either `data` (one data set) or both `assessment_data` and `training_data`.

If we choose the latter, we won't be able to run any of deepchecks's single-dataset checks.
OTOH, if we choose the former, then this snippet of code (lens.py L327-333), as presently written, will make it such that the deepchecks evaluator only ever sees ONE data set at a time and so we wouldn't be able to run any of deepchecks' multi-dataset checks.

esherman-credo: Update: Workaround 3 (which I'm going to pursue for now): implement a new `check_requirements` function that only raises a `ValidationError` if neither dataset is present

PR #215, opened 2022-10-26
Bug/210/shap error multi class

fabrizio-credo:
## Fixing Shap behavior for multiclass model

Shap values output changes when the model prediction is multiclass.
Evaluator was changed to account for that. For multiclass:

- Table output contains a class column that specifies the reference class for the shap values

## Re-alligned feature drift with Platform
- Changed name and labels structure to re-allign with platform

### DIFF
--- a/credoai/evaluators/shap.py
+++ b/credoai/evaluators/shap.py
@@ -1,3 +1,4 @@
+import enum
 from typing import Dict, List, Optional, Union
 
 from numpy import abs, mean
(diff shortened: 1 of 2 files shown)


credoai/evaluators/shap.py
```diff
@@ -104,16 +105,16 @@ def _validate_arguments(self):
...
+        ]
 
         ## Sample specific results
         if self.samples_ind:
-            labels = {"ordered_feature_names": self.shap_values.feature_names}
             ind_res = self._get_mult_sample_shapley_values()
-            self.results += [
-                TableContainer(ind_res, **self.get_container_info(labels=labels))
-            ]
+            self.results += [TableContainer(ind_res, **self.get_container_info())]
```
esherman-credo, line 117: To confirm, passing feature names to results container no longer needed because everything is handled in DataFrames directly below?

esherman-credo (approved):
Ok this looks good

PR #216, opened 2022-10-26
Feat/vocalink requirements matching

fabrizio-credo:
## Update to ensure vocalink requirements compatibility

- Updated Feature Drift evidence structure
- Updated pipeline_creator for CSI and ROC evidence requirements matching
- Updated DataProfiler name
- Slight alteration to `governance.export()`: removed output, info updated messages

### DIFF
--- a/credoai/lens/pipeline_creator.py
+++ b/credoai/lens/pipeline_creator.py
@@ -23,7 +23,7 @@ class PipelineCreator:
 
 def process_evidence_requirements(evidence_requirements: List[EvidenceRequirement]):
     evaluators = set()
-    kwargs = defaultdict(dict)
+    kwargs: dict = defaultdict(dict)
     for e in evidence_requirements:
         labels = e.label
         evaluator = labels.get("evaluator")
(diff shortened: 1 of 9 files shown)


IanAtCredo (approved):
Looks good. 

Random: I don't know what's going on, but using capitalized "FEAT" instead of "feat" is making it difficult for me to locally track your github branches.

credoai/lens/pipeline_creator.py
```diff
@@ -38,7 +39,14 @@ def process_evidence_requirements(evidence_requirements: List[EvidenceRequiremen
                 metrics.add(labels["metric_type"])
             elif "metric_types" in labels:
                 metrics = metrics.union(labels["metric_types"])
+            elif "table_name" in labels:
```
IanAtCredo, line 42: I don't think this work at all.

IanAtCredo: Gonna try and fix. This breaks if you ask for `disaggregated_table` and also doesn't work for performance, which doesn't return an `roc_curve` table_name

PR #217, opened 2022-10-27
Bugfix/generator

IanAtCredo:
## Change description

* Changed threshold-dependent evidence tables so that they specify a metric type. This was suggested by @fabrizio-credo and I agree. Table names have been made more generic
* Fixed evidence bug
* Fixed bug with metadata not being added to data profiler
* Added more data artifacts to metadata
* Added evaluator metadata to evidence metadata associated with evaluator. Also added `dataset_type` to label, if it is included in metadata
    * The way I did this was add the metadata directly to the evaluator so the evaluator could keep track of it and append it to evidence the way we append other evaluator metadata (like artifacts). This makes me think we can get rid of `PipelineStep`, the separation of metadata and evaluator, and just have the user specify the metadata they want when instantiating the evaluator. @fabrizio-credo , thoughts?


credoai/evaluators/evaluator.py
```diff
@@ -104,11 +105,20 @@ def get_container_info(self, labels: dict = None, metadata: dict = None):
...
-        save_keys = {"model": "model_name"}
+        save_keys = {
+            "model": "model_name",
+            "data": "data_name",
+            "assessment_data": "assessment_data_name",
+            "training_data": "training_data_name",
+        }
         for k in self.artifact_keys:
             save_key = save_keys.get(k, k)
             try:
```
fabrizio-credo, line 124: Just curious, in which case does this throw back an error?
`save_key` always returns something and `self.__dict__[k].name` should
as well since `k` is part of the artifact keys which are created when
we are passing the artifacts to the evaluator.

IanAtCredo: I think it was sensitive features.

credoai/lens/pipeline_creator.py
```diff
@@ -1,6 +1,6 @@
-from cProfile import label
 import inspect
 from collections import defaultdict
+from cProfile import label
```
fabrizio-credo, line 3: This is not used I think.

fabrizio-credo, line 3: This library is not used I think.

fabrizio-credo:
> This makes me think we can get rid of PipelineStep, the separation of metadata and evaluator, and just have the user specify the metadata they want when instantiating the evaluator.

I am not quite sure I am getting the question. The metadata is something we are pretty much using internally to populate the evidence with what we need for the platform to work. The user wouldn't know what to put there...unless you meant internally leveraging the evaluator metadata directly without the need for `PipelineStep`.

fabrizio-credo (approved):

fabrizio-credo:
> My understanding is that this metadata is added by a user to contextualize their evaluator in some way that is outside of configuration. @fabrizio-credo , you were the original one to add metadata to the pipeline for each evaluator, what was your thinking here?

Yes that was indeed the original thought behind it.

> Regardless of what the point is, right now it's likely unnecessary to have it supplied as a separate component in a PipelineStep, rather than just having the pipeline a list of evaluators, where the user can pass metadata straight to the evaluator.

Agreed, I had interpreted the metadata part of PipelineStep as the point in which any extra metadata supplied by the user directly at init time would be merged with any auto generated info which is necessary for the evidences.

PR #220, opened 2022-10-27
Feat/deepchecks

esherman-credo:
## Change description

Implement deepchecks evaluator and evidence +container types. By default, runs a subset of the deepchecks `model_evaluation` Suite. Capable of accepting, as input, any list of Checks objects, including checks with passing conditions specified.

`get_results()` and `get_evidence()` currently have incongruous designs. `get_results()` returns a deepchecks `SuiteResult` object which can be printed to provide the standard deepchecks interactive graphic with failed, passed, other, and not-run checks. `get_evidence()` returns a `TableEvidence` object with an underlying dataframe consisting of two columns: the name of the check, and the results (pass, fail, etc.); switching this to return `DeepchecksEvidence` (already implemented) is trivial.

Re: merge conflict/tests -- many changes were pushed to develop between the time this branch was created and now. We will need to do some serious manual merging!

## Type of change

### Development

- [ X] Application changes have been tested thoroughly


credoai/evaluators/deepchecks.py
```diff
@@ -0,0 +1,143 @@
...
+        -------
+        self
+        """
+        self._setup_deepchecks()
+
+        if self.train_dataset and self.test_dataset:
+            self.results = [
+                DeepchecksContainer(
+                    self.name,
+                    self.suite.run(
```
IanAtCredo, line 105: Though it is simple to write `self.suite.run`, there's slightly more in `evaluate` than I'd like to see. A separate function called `run_suite` would be better, and then evaluate takes the output of that and just wraps it into a `DeepchecksContainer`

credoai/evidence/containers.py
```diff
@@ -30,29 +33,32 @@ def __init__(self, evidence_class, df, labels=None, metadata=None):
...
-        return self._df
+        return self._data
 
     @abstractmethod
     def to_evidence(self):
         pass
 
+    def _validate_inputs(self, data):
+        if not isinstance(data, pd.DataFrame):
+            raise ValidationError("'df' must be a dataframe")
```
IanAtCredo, line 60:
```suggestion
            raise ValidationError("'data' must be a dataframe")
```

credoai/evidence/containers.py
```diff
@@ -30,29 +33,32 @@ def __init__(self, evidence_class, df, labels=None, metadata=None):
         ----------
         evidence_class : Evidence
             An Evidence class
-        df : pd.DataFrame
+        data : pd.DataFrame
```
IanAtCredo, line 36: No always a dataframe now. Documentation should change.

credoai/evidence/containers.py
```diff
@@ -104,11 +110,13 @@ def _validate(self, df):
 class ProfilerContainer(EvidenceContainer):
     """Container for al profiler type evidence"""
 
-    def __init__(self, data, labels: dict = None, metadata: dict = None):
-        super().__init__(ProfilerEvidence, data, labels)
+    def __init__(self, df, labels: dict = None, metadata: dict = None):
```
IanAtCredo, line 113: Why did you change this to df?

IanAtCredo (changes requested):
Minor changes requested, along with the conflicts. Very easy to read and integrate, which was what i hoped.

I'll actually run it once those changes are made and the conflicts resolved.

IanAtCredo (approved):

IanAtCredo (approved):
Not able to see the widget - any understanding of what may be causing this? Otherwise lg.

image

esherman-credo:
You're indexing differently from how I'd expect. Shouldn't `get_results()` return a dictionary?

From there you're using `results` whereas I thought this should be an auto-generated name with the evaluator name and a UUID. For instance, in my local notebook, I have to print to check first and see that the suffix is the UUID d4d0cc:

image

esherman-credo:
Update: needed to git pull locally to merge with changes to the Evaluator naming.

Now I get the right thing:
image


I suspect it's something like you don't have ipywidgets installed?

PR #228, opened 2022-11-02
Tests/run quickstart

esherman-credo:
## Change description
Add test which runs the quickstart notebook (in native python form)

Useful test to ensure overall working state. Will help ensure that new releases will run with quickstart at all times.


docs/notebooks/quickstart.ipynb
```diff
@@ -611,11 +611,11 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.12"
+   "version": "3.10.6"
```
fabrizio-credo, line 614: Just a general comment: we strive to support all versions of python from 3.8 onward.

tests/test_quickstart.py
```diff
@@ -0,0 +1,125 @@
...
+        y_train,
+        X_test,
+        y_test,
+        sensitive_features_train,
+        sensitive_features_test,
+        model,
+    )
+
+
+class Base_Evaluator_Test(ABC):
```
fabrizio-credo, line 48: Is this class necessary, since it is empty?

fabrizio-credo (approved):
Looks good. I don't think the addition of the abstract class is necessary in this instance.

Musings: I wonder if there is a way to run the notebook code content from script. That would mean that we don't need to update this test code if we make any changes to the notebook.

docs/notebooks/quickstart.ipynb
esherman-credo: I think this is just auto-changed by the conda environment I was running in (and ditto whoever ran it before me). Not sure how to force it to stay the same on Github.

tests/test_quickstart.py
esherman-credo: Removed

Re run notebook code: jupyter is not presently in our dependency list for lens. There's maybe a good reason that it shouldn't be (e.g. Lens is just a python package). We'd have to separately add the dependency specifically for the tests.

PR #232, opened 2022-11-03
moved lens to use connect

IanAtCredo:
## Change description

Removing governance functionality from Lens and moving to it's own repo


credoai/evaluators/data_profiler.py
```diff
@@ -2,11 +2,11 @@
 
 import matplotlib
 import pandas as pd
+from connect.evidence.lens_evidence import DataProfilerContainer
+
 from credoai.artifacts.data.tabular_data import TabularData
 from credoai.evaluators import Evaluator
 from credoai.evaluators.utils.validation import check_data_instance
-from credoai.evidence import ProfilerContainer
-from credoai.utils.common import ValidationError
```
esherman-credo, line 9: This is just an unused import thing?

credoai/evaluators/ranking_fairness.py
```diff
@@ -1,63 +1,183 @@
...
+            self.lb_bin = np.array(lb_bin)
+            self.ub_bin = np.array(ub_bin)
+        else:
+            self.lb_bin = lb_bin
+            self.ub_bin = ub_bin
 
     required_artifacts = ["data", "sensitive_feature"]
 
     def _setup(self):
-        self.sensitive_features = np.array(self.data.sensitive_feature)
```
esherman-credo: It's not clear exactly what these changes these are for. They track back to a commit from Amin.

I can't request changes but also can't properly review.

esherman-credo (approved):
All comments I left were about the weird commits stacking across PRs. Looks good.

esherman-credo (commented):
Tested locally with 
1. quickstart notebook (as is, without swapping to connect for governance connection)
2. governance integration notebook (as is, without swapping to connect for governance connection)
3. one of my local notebooks for testing deepchecks (as is)
4. governance integration notebook with switch to connect

Takeaways: 
- the existing Lens code (evaluators, etc.) continues to work/is not broken by this change
- connect code seems to work
- It probably makes sense to update the quickstart and governance integration docs as part of this PR to use connect.governance rather than importing Governance. If we don't, we will forget to do so until someone encounters a bug or something that doesn't work properly because they're still importing Governance the old way.

credoai/evaluators/data_profiler.py
IanAtCredo: Yeah, exactly.

IanAtCredo:
I updated the code so `from credoai.governance import Governance` still works.

esherman-credo (approved):
Looks good

PR #233, opened 2022-11-03
modify quantization of roc and pr curve interpolation helpers to ensure adequately dense interpolation

esherman-credo:
## Change description

For some applications, ROC curve appears cut off when plotted. This is due to insufficiently granular interpolation of outputs of sklearn's `roc_curve` function.

Modified quantization level in interpolator helper functions to be set as a function of 1) number of thresholds returned by `roc_curve` function (or pr curve or det curve) and 2) difference between max and min threshold.

## Type of change
- [ X] Bug fix (fixes an issue)

### DIFF
--- a/credoai/modules/credoai_metrics.py
+++ b/credoai/modules/credoai_metrics.py
@@ -233,8 +233,11 @@ def ks_statistic(y_true, y_pred) -> float:
     return ks_stat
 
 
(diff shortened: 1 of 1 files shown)


amrasekh (approved):
It works as expected and desired now.

PR #234, opened 2022-11-03
Docs/list metrics

esherman-credo:
## Change description

Modify https://credoai-lens.readthedocs.io/en/latest/metrics.html# to support call to `list_metrics` so all supported metrics can be displayed.

NOTE: I TESTED THIS "LOCALLY" IN A VSCODE RENDER OF THE HTML. I DO NOT KNOW HOW TO TEST ON READTHEDOCS WITHOUT MERGING TO DEVELOP BRANCH.

## Type of change

### DIFF
--- a/credoai/evidence/evidence.py
+++ b/credoai/evidence/evidence.py
@@ -5,7 +5,6 @@ import json
 import pprint
 from abc import ABC, abstractproperty
 from datetime import datetime
-from symbol import parameters
 from typing import Tuple
 
 from credoai.utils import ValidationError
--- a/credoai/lens/lens.py
+++ b/credoai/lens/lens.py
@@ -34,7 +34,9 @@ class PipelineStep:
         self.metadata["evaluator"] = self.evaluator.name
 
     def check_match(self, metadata):
-        """Return true if metadata is a subset of pipeline step's metadata"""
+        """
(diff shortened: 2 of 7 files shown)


fabrizio-credo (changes requested):
So I activated your version on read the docs, and tested the build. It failed, you can check the log here: 

https://readthedocs.org/projects/credoai-lens/builds/18532017/

It seems to be an issue due to import of numpy... not sure what the reason is.

MInor issue: The local html creation works fine, there are a bunch of non critical errors, somehow it tries to import ProfilerEvidence and DataProfiling, which doesn't exist anymore.
Not sure if this is critical, can you reproduce this?

image

esherman-credo:
I don't get those warnings/errors exactly. I get _a lot_ of them. And this import part looks like it's trying to import those evaluators.

image

I just looked in develop and I still see data_profiler so I'm not sure what you're referring to?


Re numpy: I suspect that's an error because numpy is imported in metric_utils (which is where `list_metrics()` is defined). I'll add that and we'll see!

esherman-credo:
Alright, after some help activating the RTD page for this branch from @fabrizio-credo, we're good to go:

https://credoai-lens.readthedocs.io/en/docs-list_metrics/metrics.html

fabrizio-credo (approved):

PR #236, opened 2022-11-07
Docs/update governance notebook

fabrizio-credo:
# Modification to notebooks

- Moving run script functionality inside lens package so notebooks are runnable by a user with standard lens installation
- Update language concerning governance in the quickstarts notebooks

### DIFF
--- a/docs/notebooks/platform_integration.ipynb
+++ b/docs/notebooks/platform_integration.ipynb
@@ -17,7 +17,7 @@
     "\n",
     "**Find the code**\n",
     "\n",
-    "Download this notebook from [github](https://github.com/credo-ai/credoai_lens/blob/develop/docs/notebooks/platform_integration.ipynb).\n",
+    "Download this notebook from [github](https://github.com/credo-ai/credoai_lens/blob/main/docs/notebooks/platform_integration.ipynb).\n",
     "\n",
     "If you want to just get the boilerplate code, see [example in this script](https://raw.githubusercontent.com/credo-ai/credoai_lens/develop/docs/example_code.py)."
    ]
--- a/docs/notebooks/platform_integration.ipynb
+++ b/docs/notebooks/platform_integration.ipynb
@@ -96,7 +96,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Governance in 5 Minutes"
+    "## Platform integration in 5 Minutes"
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


esherman-credo (changes requested):

fabrizio-credo:
General question: Is there a need for tags always? Tags is a functionality, but it's not mandatory. Shouldn't the example be the most general?

esherman-credo:
Posting this screenshot of a traceback I had previously shared directly with @fabrizio-credo:

image


I guess tags maybe shouldn't be the default in the integration walkthrough at the top of the notebook. But it definitely needs to be mentioned somewhere in this notebook. Maybe a section at the bottom that's like "here's this optional feature we recommend"? I didn't understand the error message at first and I'm internal/aware of the point of tags.

I think the confusion comes from the fact that the traceback arises when you call `Lens.run`, which is rather detached from where the issue actually starts. The issue is caused by not specifying an optional argument (which, in this case, is not optional based on the details of other optional-but-supplied arguments) when calling the `ClassificationModel` constructor. That issue is propagated to the Lens constructor (issue propagated silently) and then doesn't pop up until we call `run()`. The `WARNING` message is helpful but probably not sufficient because most peoples' instinct is going to be to follow the actual error message.

I think it might be valuable to do this validation in the Lens constructor, since that's where the conflict arises. cc @IanAtCredo 

It might be possible to do it even earlier, at the time we wrap the model as a `ClassificationModel` (or regression model), but I'm less sure about that since the issue is caused by the intersection of the model specification and the evidence requirements coming in from Platform

fabrizio-credo:
@esherman-credo yeah I will add a section about tags in the notebook. We can revisit the issue of the error when the new connect package is redefined. As you say, this is an issue about the intersection between model config and requirements. Note: there can still be reasons to run models without tags even though there are requirements with tags, we need to keep that in mind.

docs/notebooks/platform_integration.ipynb
```diff
@@ -26,10 +26,9 @@
...
    "source": [
-    "## Glossary\n",
-    "A list of useful keywords:  \n",
+    "## Useful keywords\n",
     "\n",
-    "- **Credo AI Platform**: Also referred to as simply \"platform\". The central AI governance/reporting Platform, found at [https://app.credo.ai/](https://app.credo.ai/)\n",
+    "- **Credo AI Platform**: Also referred to as simply \"Platform\". The central AI governance/reporting Platform, found at [https://app.credo.ai/](https://app.credo.ai/)\n",
     "\n",
     "- **credoconfig**: configuration file to be copied in the user's home folder\n",
     "\n",
```
esherman-credo, line 34: Line 39 -> "(where do we get this?)

Either answer that question or delete (ideally the first)

esherman-credo (approved):
Please address comment but otherwise good.

PR #247, opened 2022-11-14
Feat/black workflow checker (WIP)

IanAtCredo:
## Change description
Added black format checker to PR upload and reformated


docs/setup.rst
```diff
@@ -22,6 +22,14 @@ The latest development version can also be installed directly from github.
 
    pip install git+https://github.com/credo-ai/credoai_lens#develop
 
+If you are interested in contributing to the development of Lens and
+would like to install requirements for testing and formatting, you'll
+have to install the dev requirements.
+
+::
+
+   pip install credoai-lens[dev]
```
esherman-credo, line 31: How will this impact using `pip install -e .` for getting the installed version of Lens to change depending on git branch?

esherman-credo (approved):
1 inline question regarding lens install via git folder

Another question: at what point should we update python requirements to something >3.8?

Neither is obstacle to merging.

docs/setup.rst
IanAtCredo: It won't change anything.  If you want to install locally with all dependencies you'd type `pip instlal -e ".[dev]"`. But that just affects the installation of dependencies. What you are talking about is referencing the local (changing) folder for the actual credoai library, which will be unchanged.

PR #248, opened 2022-11-15
Feat/prism

fabrizio-credo:
### Overall folder/file structure

- **comparators**: Contains the basic logic to compare different type of results. Comparators can be a function of the result container type, and potentially the evaluators
  - _comparator.py_: abstract class
  - _metric_comparator.py_: takes a list of Lens objects, extract metric containers and compares metrics across all the Lens objects. Comparisons are done respect to a reference Lens object. Currently the reference is defined by model name, to be changed in the future if necessary. @esherman-credo I simplified the logic, the all vs all comparison produced a lot of results that were probably not going to get used, we can reinstate it if needs be. I have added the possibility to select different operations.
- _compare.py_: this is a type of _task_ that prism can orchestrate. ATM it's only calling metric_comparator, in the future `compare.Compare` will instantiate the necessary comparators (table, metric, etc...) depending on the existing results containers.  

- _prism_.py: the orchestrator, this is the equivalent of Lens but at a higher level. The purpose of **Prism** is to coordinate the various **tasks**.  Currently when the user instantiates it, they can pass a dictionary of tasks to perform on the Lens objects. The dictionary is in the format `{'task_name:{'parameter':value}}`, where parameters are specific to the type of task to perform.

### Missing

- Docstrings in general
- Abstract classes for task objects (like `compare.Compare`). I haven't created an abstract class for this yet, I am wondering if there is a need for it, or if the variety of the object will be too big for an abstract class to be useful. 
- Tests

To test the run you can use the following code:

```

from credoai.lens import Lens
from credoai.artifacts import TabularData, ClassificationModel
from credoai.evaluators import *
from credoai.governance import Governance
import numpy as np


# imports for example data and model training
from credoai.datasets import fetch_creditdefault
from sklearn.ensemble import RandomForestClassifier
from sklearn.model_selection import train_test_split
from xgboost import XGBClassifier

data = fetch_creditdefault()
df = data["data"]
df["target"] = data["target"].astype(int)

# fit model
model1 = RandomForestClassifier(random_state=42)
X = df.drop(columns=["SEX", "target"])
y = df["target"]
sensitive_features = df["SEX"]
(
    X_train,
    X_test,
    y_train,
    y_test,
    sensitive_features_train,
    sensitive_features_test,
) = train_test_split(X, y, sensitive_features, random_state=42)
model1.fit(X_train, y_train)

model2 = XGBClassifier(tree_method='hist', random_state=42, n_estimators=2, enable_categorical=True)
model2.fit(X_train, y_train)



# ## Set up artifacts

credo_RF1 = ClassificationModel(
    'credit_RF1_classifier',
    model1
)

credo_XGB = ClassificationModel(
    'credit_XGB_classifier',
    model2
)

train_data = TabularData(
    name = "UCI-credit-default-train",
    X=X_train,
    y=y_train,
    sensitive_features=sensitive_features_train
)
test_data = TabularData(
    name='UCI-credit-default-test',
    X=X_test,
    y=y_test,
    sensitive_features=sensitive_features_test
)


# pipeline scan be specifed using a sklearn-like style
metrics = ['accuracy_score', 'roc_auc_score']
pipeline = [
    (Performance(metrics)),
    (ModelFairness(metrics)),
]

pipeline2 = [
    (Performance(metrics)),
    (ModelFairness(metrics)),
]


# and added to lens in one fell swoop
lens_rf1 = Lens(
    model=credo_RF1,
    assessment_data=test_data,
    # training_data=train_data,
    pipeline=pipeline
)



lens_xgb = Lens(
    model=credo_XGB,
    assessment_data=test_data,
    # training_data=train_data,
    pipeline=pipeline2
)


from credoai.prism import Prism

prism_test = Prism(
    [lens_rf1, lens_xgb], 
    tasks={'compare':{'ref':'credit_RF1_classifier'}}
)

prism_test.execute()

prism_test.get_results()


```


credoai/prism/comparators/metric_comparator.py
```diff
@@ -0,0 +1,106 @@
...
+            Columns:
+
+                type: metric name
+                value: original value of the metric
+                id: Identifier of the origin of the metric
+                comparison: value of the comparison
+        """
+        # Group by metrics and calculate comparison
+        output = []
+        for _, results in self.all_results.groupby("type"):
```
IanAtCredo, line 95: What happens if there are multiple metrics of the same type? There are at least N-model metrics, but there may be other dimensions? No metrics containers are like this currently, but they _can_ have more dimensions.

For instance, if we calculated "disaggregated performance" using a list of metrics instead of a table, we'd have "sensitive_feature" as a column. We'd want to compare apples-to-apples - find the reference model that has the same row (except for the value) and compare it to the other.

credoai/prism/comparators/metric_comparator.py
```diff
@@ -0,0 +1,106 @@
...
+            ref_value = results.value.loc[results.id == self.overall_ref].iloc[0]
+            if ref_value:
+                results["comparison"] = self.operation(results.value, ref_value)
```
IanAtCredo, line 98: So each result will have the reference compared to itself

credoai/prism/compare.py
```diff
@@ -0,0 +1,64 @@
...
+    ):
+        """
+        Compare results between 2 different Lens runs.
+
+        Depending on the type of containers detected, this class will invoke
+        the appropriate comparator.
+
+        Parameters
+        ----------
+        results_primary : List[EvidenceContainer]
```
IanAtCredo, line 29: These parameters don't match the initialization. Also the docstring is under __init__ rather than the class name

credoai/prism/prism.py
```diff
@@ -0,0 +1,49 @@
+from typing import List
+from credoai.lens import Lens
+from credoai.utils import ValidationError, global_logger
+from credoai.utils.common import NotRunError
+from credoai.prism.compare import Compare
+
+
+class Prism:
+    SUPPORTED_TASKS = {"compare": Compare}
+
```
IanAtCredo, line 10: documentation needed

credoai/prism/comparators/metric_comparator.py
```diff
@@ -0,0 +1,106 @@
...
+        Runs all comparisons
+        """
+        # Calculate scalar differences across all the metrics
+        self._scalar_operation()
+        return self
+
+    def _extract_results_from_containers(self):
+        # Assign model name as ID
+        self.all_results = [
+            res.df.assign(id=res.metadata["model_name"])
```
IanAtCredo, line 74: Is model name really enough to separate results? What about evidence that is the same model calculated on different datasets?

credoai/prism/compare.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import List, Literal, Optional
+from credoai.evaluators.utils.validation import check_instance
+from credoai.evidence.containers import MetricContainer
+from credoai.lens import Lens
+
+from credoai.utils import flatten_list
+from credoai.prism.comparators.metric_comparator import MetricComparator
+
+
+class Compare:
```
IanAtCredo, line 11: Why does this class need to exist? I thought this class was essentially what Prism was doing

credoai/prism/prism.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self, network: List[Lens], tasks: dict):
+        self.network = network
```
IanAtCredo, line 12: The usage of this class is pretty opaque.

`Network` is a word that comes out of nowhere. This is a list of lens objects? Do we also want to support lists of evidence containers here? Seems a bit restricted in scope while also not being that transparent.

`Tasks` is very opaque. I feel like we should automatically be able to run comparators, with optional parameterization. I also feel like you _should not_ have to define reference class by string, but instead by model.

credoai/prism/comparators/metric_comparator.py
```diff
@@ -0,0 +1,106 @@
...
+        "diff": lambda x, ref: x - ref,
+        "ratio": lambda x, ref: x / ref,
+        "perc": lambda x, ref: x * 100 / ref,
+        "perc_diff": lambda x, ref: ((x - ref) / ref) * 100,
+    }
+
+    def __init__(
+        self,
+        EvidenceContainers: List[MetricContainer],
+        ref: Optional[str] = None,
```
IanAtCredo, line 38: What happens if there is no `ref`? Do we want to default to just the first evidencecontainer?

IanAtCredo (changes requested):
Left some comments. Bullet point high level critiques:

* I feel like the `compare.py` class is unneeded and can be rolled into Prism. Happy to discuss and be wrong here though.
* The way a reference is defined and IDed is not flexible enough. Only works for cross-model comparisons, when really we want cross "context" comparisons. If the same model is run on different samples of data, we may want to compare those.
* The user interface of Prism, which is supposed to simplify everything, instead obfuscates. 
* Documentation needed for most functions (though easy to follow anyway)

fabrizio-credo:
@IanAtCredo Overall responses to the high level critique:

> I feel like the compare.py class is unneeded and can be rolled into Prism. Happy to discuss and be wrong here though.

Initially I was of the same opinion, however I feel we need this kind of objects due to the variety of tasks that **Prism** will have to handle in the future (not just comparisons). If everything gets put into **Prism** I feel it will eventually ballon up pretty quickly. I intended for Prism to only handle orchestration logic, including the translation from user input into the tasks to perform.
This is based mostly on feeling at the moment, I would like to create a sampling _task_ following this schema and see how we feel about it at the end.


> The way a reference is defined and IDed is not flexible enough. Only works for cross-model comparisons, when really we want cross "context" comparisons. If the same model is run on different samples of data, we may want to compare those.

Agreed,  this was meant as a temporary solution. I think Data/Model/SensitiveFeature define the context ATM, I wonder what is the best way of conveying this information. The simplest is a combo of Model_Name + Data_Name + Feature_Name/Value, this will get ugly pretty fast.

Potential solution: leverage `PepelineStep` -> Create a `context` property which is a function of metadata, and which associate a hash to the step. In my mind each step coincides with a context.  All Prism _tasks_ will use hashes for internal referencing (this makes creating DF easier, rather than having to move around Dictionaries).
We can create a couple of Prism methods that can then translate to/from human readable output/input.

> The user interface of Prism, which is supposed to simplify everything, instead obfuscates.
> Documentation needed for most functions (though easy to follow anyway)

Agreed, I wasn't extremely happy about it either. We can ask the user to pass the actual (task) class, like `Compare`. So the param `tasks` in Prism will be something like this:

```
[
 {
   task: Compare,
   params: {ref: ...}
 },
]
```
another argument to keep the _task_ objects 😛 . In any case, whether we map user input string to behavior, or we pass tasks directly, a certain amount of knowledge is necessary on user side. 
Stylistically this is more similar to what we are currently doing in Lens when we build the pipeline.

IanAtCredo:
Few thoughts:

Why does the user need to specify tasks _at all_? Or at the level of granularity you are describing? Why not just "here are a few models, either aggregate or compare them". I can buy that if you say "aggregate them", _implicitly_ a list of tasks are created, but should that be the main way the user interacts? Perhaps some super-user will want specific access to the tasks (which we can allow), but that's secondary to the primary user interaction.

Data/Model/SensitiveFeature as context - yeah. I think we may want to define this somewhere. I was originally thinking that a _run of Lens_ is the context. Why do we need to infer the context? Can't we just say "we took a bucket of what you said was one thing and compared it to another." To generalize this to evidence separated from Lens is trickier within your current implementation, but could be done by a list of evidence-lists, rather than inferring separation by IDs.

fabrizio-credo:
@IanAtCredo Update:

- Formalized the concept of Task
- Updated method of passing tasks to Prism, same logic as for passing evaluators to Prism
- Updated PipelineStep to create a unique id for each step dependent on Evaluator name, model name, datasets names, sensitive feature.
- Update comparator task and metric comparators to leverage the new id to match evidences for comparisons

credoai/prism/comparators/metric_comparator.py
```diff
@@ -0,0 +1,106 @@
+"""Comparators for metric type containers"""
+from typing import Callable, List, Literal, Optional
+
+import numpy as np
+from pandas import DataFrame, concat
```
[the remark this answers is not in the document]
fabrizio-credo: Interestingly I always thought referencing only the method you need was preferable, mainly based on the Zen of Python: "Explicit is better than implicit". Also if the imported package is big, it makes evident which are the parts we depend on.

Why was your preference for the `import package as ...` syntax ?

credoai/lens/lens.py
```diff
@@ -35,6 +36,27 @@ def __post_init__(self):
         self.evaluator.metadata = self.metadata
         self.metadata["evaluator"] = self.evaluator.name
 
+    @property
+    def identifier(self):
```
IanAtCredo, line 40: I spent a while trying to make this more robust. As it is, this seems quite brittle! We're passing an ID that is made up on-the-fly here, to be referenced via a hardcoded set of ID_Cols. Doesn't seem good.

Tried to replace with this:
```
    @property
    def id(self):
        eval_info = self.evaluator.get_info()["metadata"]
        return "~".join(f"{k}:{v}" for k, v in eval_info.items())
```
which makes use of another function we already have to create metadata (I also changed "identifier" => "id")

This results in outputs like this:
`'evaluator:Performance~model_name:credit_RF1_classifier~assessment_dataset_name:UCI-credit-default-test'`

This can't be used later however because the metadata isn't consistent amongst evaluators. However, I'm confused why that _should_ be an issue. Shouldn't we be able to do comparisons evaluator-by-evaluator when they match up? The issue now is coming up because everything is being coerced into one dataframe.

I tried to remove the explicit separation of ID into ID_Cols, but then saw that the ID_Cols were used so that evaluation could be directed to be over the `ref class`, like "model". I was confused about that implementation as well. Once we know that we have the same ID between two lens instances, shouldn't we be able to compare? Why do we need a ref class at all?

credoai/prism/prism.py
```diff
@@ -0,0 +1,86 @@
...
+    an arbitrary amount of Lens objects.
+
+    Parameters
+    ----------
+    lenses : List[Lens]
+        A list of Lens objects. The only requirement is for the Lens objects to
+        be instantiated with their necessary artifacts. One or multiple Lens objects
+        can be provided, each Task will validate that the amount of objects provided
+        is suitable for its requirement.
+    task : Task
```
IanAtCredo, line 23: You said tasks work as evaluators, but here I see I can only pass one task.

credoai/prism/comparators/metric_comparator.py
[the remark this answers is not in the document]
IanAtCredo: My preference for `import package as` is because I find it _more explicit_. We import the entire package regardless and you can easily see that this is "pd.DataFrame", rather than some other DataFrame. It's also the convention I see everywhere with pandas and numpy.

IanAtCredo (commented):
I'm not sure if changes need to be made, but I feel like the current "identifier" implementation is quite shaky and will lead to problems down the road. Let's discuss whether changes are necessary. 

We can also call this an experimental feature, accept some brittleness and see if it is required by others.

credoai/lens/lens.py
fabrizio-credo: The reference was split in `ref_type` and `ref` to allow the flexibility of comparing either by model or by whichever dataset (training/assessment) is available. This is to accommodate 2 type of use cases:

1. Many models comparisons: multiple iterations of the same model vs initial case, or different models against a reference one.
2. Multiple dataset comparisons: comparing result when dataset change, over time or due to some form of sampling for example.

If Metadata structure was always constant then I would agree that we can simply pass the `self.evaluator.get_container_info()`, but as it stands I would have to still do extra steps to standardize the output in a similar way to what I am already doing. 

Standardization is also good for results displaying, I think breaking down the id in labelled columns helps user understanding the results, rather then having to read a long and variable id.

credoai/prism/prism.py
fabrizio-credo: Yeah I consciously constrained the power of Prism, task as we imagined them would be of 2 types:

1. Calculations on Lens results, produce numerical results (compare for example)
2. Operations on the instances, for example creating new ones based on sampling of datasets

I imagine that at the end of a Prism run some form of numerical results need to be produced. 
That being the case, a Type 2 task would require another task of type 1 to be run after, so that numerical output can be produced. This would require to enforce order of execution (or to add task dependency on other tasks).

I decided to avoid the complication by allowing for one task to call another one. So for example an hypothetical cross-validation task would:

1. Create the Lens instances by splitting dataset appropriately
2. Run `compare` to produce numerical results

ATM I don't imagine the need to run super complex sequences of tasks, this might be a reflection on the fact that we don't have an actual use case for Prism yet.

IanAtCredo (approved):
As mentioned, I really don't think this ID implementation is great. Or, at least, the ID specification conflicts with other "sources of truth" about artifacts reflected in `get_container_info`. I also don't think parsing an ID string is reasonable - why not query the evaluator directly during the compare step?

All that said, I don't think we should invest more time in this. It works, let's move on.

PR #251, opened 2022-11-16
Add joblib support to Lens to enable running evaluators in parallel

esherman-credo:
## Change description

Modify Lens class to support optional use of joblib.

Lens constructor now takes as optional input an `n_jobs` parameter. Parameter is used in the `Lens.run()` function to parallelize running of evaluators.

Results returned by each evaluator's `evaluate` function are evaluators, rather than pipeline steps. Therefore, results must be stored in an intermediate list and then re-matched with metadata in `self.pipeline`

## Type of change
- [x] Bug fix (fixes an issue)

### DIFF
--- a/credoai/lens/lens.py
+++ b/credoai/lens/lens.py
@@ -8,6 +8,7 @@ from inspect import isclass
 from typing import Dict, List, Optional, Tuple, Union
 
 from connect.governance import Governance
+from joblib import Parallel, delayed
 
 from credoai.artifacts import Data, Model
 from credoai.evaluators.evaluator import Evaluator
@@ -50,6 +51,7 @@ class Lens:
         training_data: Data = None,
         pipeline: Pipeline = None,
         governance: Governance = None,
+        n_jobs: int = 1,
     ) -> None:
         """
         Initializer for the Lens class.
@@ -73,6 +75,10 @@ class Lens:
             Lens and the Credo AI Platform. Specifically, evidence requirements taken from
             policy packs defined on the platform will configure Lens, and evidence created by
             Lens can be exported to the platform.
+        n_jobs : integer, optional
+            Number of evaluator jobs to run in parallel.
+            Uses joblib Parallel construct with multiprocessing backend.
+            Specifying n_jobs = -1 will use all available processors.
         """
         self.model = model
         self.assessment_data = assessment_data
@@ -85,6 +91,7 @@ class Lens:
             self.sens_feat_names = list(self.assessment_data.sensitive_features)
         else:
             self.sens_feat_names = []
+        self.n_jobs = n_jobs
         self._add_governance(governance)
         self._generate_pipeline(pipeline)
         # Can  pass pipeline directly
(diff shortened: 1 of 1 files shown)


credoai/lens/lens.py
```diff
@@ -172,9 +179,16 @@ def run(self):
...
+
+        # Run evaluators in parallel. Shared object (self.pipeline) necessitates writing
+        # results to intermediate object evaluator_results
+        evaluator_results = Parallel(n_jobs=self.n_jobs, verbose=100)(
+            delayed(step.evaluator.evaluate)() for step in self.pipeline
+        )
+
+        # Write intermediate evaluator results back into self.pipeline for later processing
+        for idx, evaluator in enumerate(evaluator_results):
+            self.pipeline[idx].evaluator = evaluator
```
fabrizio-credo, line 191: I don't know exactly how `joblib` works, but this assumes that `evaluator_results` is always in the same order as the pipeline steps. Is that always the case?

If the second in the pipeline finishes earlier than the first, does joblib waits to populate evaluator_results?

fabrizio-credo (commented):
It looks like a nifty change.

I am not familiar with how joblib works, and would like clarity on the comment I made below. I am always quite weary when we overwrite the internal state of objects.

credoai/lens/lens.py
esherman-credo: So I had tested this a little bit before pushing, but now I've done some more research to verify.

1. Wrote a function "delay_sqrt(i)" which calls time.sleep(2) if i is odd (no sleep if even) and tested --> results in order
2. Added a sleep(10) call to some of the evaluators --> results still in order
3. Per [this](https://stackoverflow.com/questions/56659294/does-joblib-parallel-keep-the-original-order-of-data-passed), multiprocessing (which I'm not using) guarantees that results are in order...I don't see that explicitly in the joblib docs. Going to [multiprocessing.Pool](https://docs.python.org/3.5/library/multiprocessing.html#multiprocessing.pool.Pool.map), the map function is what's being called and map guarantees return order, but not execution order (which is fine, but maybe not speed optimal)
4. Loky backend (which we are using) -> [implements a map function ](https://github.com/joblib/joblib/externals/loky/process_executor.py) based on the multiprocessing map function linked above (so parallel execution, finish order not guaranteed, return order is guaranteed). If you click on that last link to the loky definition, you can see that it's actually straight up calling the concurrent.futures map, which is based on vanilla Python [map](https://www.geeksforgeeks.org/python-map-function/). My interpretation of that is that under the hood what happens is i) lay out a result vector whose members are directly linked to function + argument pairs, ii) spin out the function + argument pairs to run in parallel using different processes, iii) as they return, the process is placed in a specific spot in the result list via the PID of that process (see [this](https://github.com/joblib/joblib/blob/master/joblib/externals/loky/process_executor.py#L670) function), and then iv) when everything is done all the processes clean-up and you have your in-order list according to order of arguments, not necessarily order of execution.

fabrizio-credo: Sounds great! Excellent documentation!

fabrizio-credo (approved):

PR #252, opened 2022-11-17
Add functionality to check version of Lens on import

esherman-credo:
## Change description
Changes credoai-lens's version to be based on installed wheel, rather than hardcoded string.

Adds functionality to check latest version of package on PyPI. If installed version is not identical to latest version, Lens's logger will throw a warning instructing the user to update.

## Type of change
- [ ] Bug fix (fixes an issue)

### DIFF
--- a/credoai/__init__.py
+++ b/credoai/__init__.py
@@ -2,4 +2,14 @@
 Primary interface for Credo AI Lens package
 """
 
-__version__ = "1.1.2"
+from importlib.metadata import version
+
+# get package version programatically from installed wheel
+# does not rely on update to __init__.py for each new release
+# __version__ = version("credoai-lens")
+__version__ = "1.0.0"
+
+
+from credoai.utils.version_check import validate_version
+
+validate_version()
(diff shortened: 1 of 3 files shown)


credoai/__init__.py
```diff
@@ -2,4 +2,14 @@
 Primary interface for Credo AI Lens package
 """
 
-__version__ = "1.1.2"
+from importlib.metadata import version
+
+# get package version programatically from installed wheel
+# does not rely on update to __init__.py for each new release
+# __version__ = version("credoai-lens")
+__version__ = "1.0.0"
```
IanAtCredo, line 10: Why are you removing this version? This is the source of truth for what version the package is. `Setup.py` refers to this, as does pypi, through setup.

IanAtCredo (changes requested):
The validation looks good. The change to `__init__.py` doesn't make sense to me. If I'm missing something, please explain - but `__init__.py` is where we define the version.

credoai/__init__.py
esherman-credo: The commented part was pattern matching based on deepchecks and doesn't work for us (hence commented)

The __version__ = "1.0.0" part was me testing to see that the warning would output if I imported the package locally.

Both were committed by mistake.

IanAtCredo (approved):

PR #257, opened 2022-11-23
Tests/export json

esherman-credo:
## Change description

Add calls to `governance.export()` to each test via internal method `governance._file_export()`. Need to use this helper function since tests do not register governance to a Platform use case/policy pack/api url.

Helps ensure that JSON writing is successful (catches recent bugs related to datetimes and [NaNs](https://github.com/credo-ai/credoai_connect/pull/3) in JSONs).

NOTE: This is a branch off of https://github.com/credo-ai/credoai_lens/pull/256. I created it as a separate branch/PR since it's a separate feature, potentially with it's own discussion of pros/cons.

NOTE 2: This currently fails! That is because it is catching a JSON bug in connect that has not yet been merged. See https://github.com/credo-ai/credoai_connect/pull/3

## Type of change
- [x] New feature (adds functionality)

### DIFF
--- a/tests/test_lens.py
+++ b/tests/test_lens.py
@@ -28,6 +28,8 @@ from credoai.lens import Lens
 
 from connect.governance import Governance
 
+import tempfile
+
 TEST_METRICS = [
     ["false_negative_rate"],
     ["average_precision_score"],
@@ -93,6 +95,10 @@ def test_model_fairness(
     assert lens.get_results()
     assert lens.get_evidence()
     assert lens.send_to_governance()
+    tfile = tempfile.NamedTemporaryFile(delete=False)
+    assert not gov._file_export(
+        tfile.name
+    )  # governance file IO returns None if successful
 
 
 def test_privacy(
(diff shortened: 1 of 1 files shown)


IanAtCredo:
I'm not fully following the logic here. Not _everything_ should have to be tested in the same area. Is the idea that this ensures that the evaluators => containers => json work well?

esherman-credo:
Yes.

I think there's a case to be made for instantiating container/evidence objects and calling `export()` to unit test the JSON writing functionality. That sort of test should live in the connect test suite.

I think that's liable to under-cover relative to having tests which start from the evaluator and programatically generate the container/evidence objects. As an example, the NaN issue would likely have been missed by a unit test; our Container object would probably have just contained a well-formed DataFrame without NaNs. It's an edge-case that we should anticipate and write a separate test for. Nevertheless, it would be easy to miss.

This (#257) PR is, at worst, redundant. I think it's less of a set of unit tests than it is a set of integration tests. To be clear, if `run()` works for a given evaluator but `export()` does not, that will still show up in the test results. The overall test for the evaluator will be marked as failed but it will be easy to tell that the problem is not with `run()`.

fabrizio-credo (approved):
Seems fine to me.

@esherman-credo This is failing due to changes still to be released on the connect side, correct?

esherman-credo:
@IanAtCredo Waiting to merge until you have a chance to follow-up comment.

The new tests here now pass, since the workflow is uses the updated connect v0.0.4, which fixes https://github.com/credo-ai/credoai_connect/pull/3

PR #260, opened 2022-11-28
Create pull_request_template.md

IanAtCredo:
## Change description

> Description here

## Type of change
- [ ] Bug fix (fixes an issue)
- [ ] New feature (adds functionality)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally
- [ ] Application changes have been tested thoroughly
- [ ] Automated tests covering modified code pass

### Security

- [ ] Security impact of change has been considered
- [ ] Code follows company security practices and guidelines

### Code review 

- [ ] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached as necessary
- [ ] "Ready for review" label attached and reviewers assigned
- [ ] Changes have been reviewed by at least one other contributor
- [ ] Pull request linked to task tracker where applicable

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,8 +2,13 @@
 
 ## Issue ticket number and link
 
+## Known outstanding issues that are not fully accounted for
+
 ## Checklist before requesting a review
 - [ ] I have performed a self-review of my code
-- [ ] I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API documentation is constructed correctly
-- [ ] I have built tests for new functionality (particularly new evaluators)
+- [ ] I have built basic tests for new functionality (particularly new evaluators)
+- [ ] If new libraries have been added, I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API documentation is constructed correctly
 - [ ] Will this be part of a major product update? If yes, please write one phrase about this update.
+
+## Extra-mile Checklist
+- [ ] I have thought expansively about edge cases and written tests for them
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,9 @@
+## Describe your changes
+
+## Issue ticket number and link
+
+## Checklist before requesting a review
+- [ ] I have performed a self-review of my code
+- [ ] I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API still is constructed correctly
+- [ ] I have built tests for new functionality (particularly new evaluators)
+- [ ] Will this be part of a major product update? If yes, please write one phrase about this update.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,6 +4,6 @@
 
 ## Checklist before requesting a review
 - [ ] I have performed a self-review of my code
-- [ ] I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API still is constructed correctly
+- [ ] I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API documentation is constructed correctly
 - [ ] I have built tests for new functionality (particularly new evaluators)
 - [ ] Will this be part of a major product update? If yes, please write one phrase about this update.


.github/pull_request_template.md
```diff
@@ -4,6 +4,6 @@
 
 ## Checklist before requesting a review
```
esherman-credo, line 5: A general question is whether this can be made part of the template for the _reviewer_ in addition to it being a template for the PR-submitter

.github/pull_request_template.md
```diff
@@ -4,6 +4,6 @@
...
 - [ ] I have performed a self-review of my code
-- [ ] I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API still is constructed correctly
+- [ ] I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API documentation is constructed correctly
```
esherman-credo, line 7: This is sorta complicated since it will require activating the present branch on RTD and compiling. Will RTD auto-delete docs for deleted branches after they're merged?

If that's the process we want to follow, we should be explicit.

.github/pull_request_template.md
```diff
@@ -4,6 +4,6 @@
 
 ## Checklist before requesting a review
 - [ ] I have performed a self-review of my code
```
esherman-credo, line 6: Need to also mention checking against the current develop branch of connect (or modify testing workflow to run against the develop branch of connect, perhaps?)

.github/pull_request_template.md
```diff
@@ -4,6 +4,6 @@
...
-- [ ] I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API still is constructed correctly
+- [ ] I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API documentation is constructed correctly
 - [ ] I have built tests for new functionality (particularly new evaluators)
```
esherman-credo, line 8: I think we should try to be more comprehensive about what tests we write. We can push for that a bit in this review checklist.

Right now I think our internal default expectation is to add a `add()` and a `run()` call to the test suite. For evaluators we should also do `to_evidence()`, `send_to_governance()`, and `export()`; all of those can be put in a testing checklist for the PR.
- It might also be good to have the person who opens the PR to write up a summary of edge cases they thought of and wrote test cases for. This will be good for forcing both the PR-opener and the reviewer to think more expansively about ways things could go wrong. I think the default right now is to primarily try to assess whether the code will solve exactly the narrow issue it's designed to.

esherman-credo (changes requested):
Request some more detail on a few points. Some items are perhaps better suited to separate Jira tickets, like checking that code works with the connect develop branch.

.github/pull_request_template.md
IanAtCredo: I think Lens should always be run against current develop version of Connect. A release QA should make sure Connect and Lens stable versions are working together

.github/pull_request_template.md
IanAtCredo: This is more expansive and test-heavy than I think we need to be right now.

.github/pull_request_template.md
IanAtCredo: We should update the requirements of Lens to reflect that. The developer of Lens should not _need_ to be testing against different versions of Connect. Just whatever version Lens requires.

IanAtCredo:
Added one "extra-mile" check, and some extra space. I don't think we need to stress about this too much. Having a functional PR template is much better than what we had before.

.github/pull_request_template.md
esherman-credo: I'm concerned this is going to lead to review blocks between the two packages.
Say some bug in Connect causes a Lens test to fail (something with Container constructors, for example, since those are used by evaluators).
In that case the Lens PR might be "correct" per visual review and local testing with the Connect bugfix branch, but can't be merged (and probably shouldn't be approved) until the Connect bugfix is merged.

This seems like it's outside the scope of the present PR, but it's definitely something to keep on our radar.

esherman-credo (approved):
See comment regarding cross-dependence with Lens. Outside the scope of this PR, but should be on our radar to avoid development delays due to separate review processes between the two repos.

PR #266, opened 2022-11-30
Bugfix/label subtypes

### LINKED ISSUE #265: Lens labels are non-unique for some evaluators, where evidence is further differentiated by subtypes in metadata
## Expected Behavior
When Lens is matching evidences to requirements, it checks the label of the Lens-generate evidence against the label specified by a policy pack. Lens' labeling should, theoretically, be unique such that a proper labeling in the policy pack will correspond to Lens outputs.

## Actual Behavior
Evaluators output unique evidence, however sometimes the uniqueness is encoded in evidence metadata rather than in the label directly. This causes Lens error messages saying multiple evidences satisfy a given requirement.

For example, consider the following error message:
image

This corresponds to the following evidences in the outputted JSON file:
```
{
          "type": "metric",
          "label": {
            "metric_type": "MembershipInference",
            "evaluator": "Privacy"
          },
          "data": {
            "value": 0.5922,
            "confidence_interval": null,
            "confidence_level": null
          },
          "generated_at": "2022-11-30T17:01:02.771421",
          "metadata": {
            "subtype": "BlackBoxRuleBased",
            "evaluator": "Privacy",
            "model_name": "credit_default_classifier",
            "assessment_dataset_name": "UCI-credit-test",
            "training_dataset_name": "UCI-credit-train"
          }
        },
        {
          "type": "metric",
          "label": {
            "metric_type": "MembershipInference",
            "evaluator": "Privacy"
          },
          "data": {
            "value": 0.5674666666666667,
            "confidence_interval": null,
            "confidence_level": null
          },
          "generated_at": "2022-11-30T17:01:02.771446",
          "metadata": {
            "subtype": "BlackBox",
            "evaluator": "Privacy",
            "model_name": "credit_default_classifier",
            "assessment_dataset_name": "UCI-credit-test",
            "training_dataset_name": "UCI-credit-train"
          }
        },
        {
          "type": "metric",
          "label": {
            "metric_type": "MembershipInference",
            "evaluator": "Privacy"
          },
          "data": {
            "value": 0.5922,
            "confidence_interval": null,
            "confidence_level": null
          },
          "generated_at": "2022-11-30T17:01:02.771467",
          "metadata": {
            "subtype": "attack_score",
            "evaluator": "Privacy",
            "model_name": "credit_default_classifier",
            "assessment_dataset_name": "UCI-credit-test",
            "training_dataset_name": "UCI-credit-train"
          }
        },
```

The `subtype` field in the metadata is sufficient to uniquely identify the 3 different evidences, while the information in the `label` field is not. (It's not clear that it will ALWAYS be the case that looking at `subtype` + label is enough to identify, there might be other fields...)

This appears to be an issue for the Privacy and DataFairness evaluators since these inherently calculate several metrics by default and they're stored in results dataframes where there is a column for the `subtype`.

Potential solutions:
1. Add subtype to the label by pulling from the metadata object (where it currently lives) here https://github.com/credo-ai/credoai_connect/blob/develop/connect/evidence/evidence.py#L95 <- Bad solution since this is inconsistent across evaluators...unless we want the entire label to be [present label] (union) [metadata]...still probably a bad idea.
2. Pass extra labeling information to the evidence object when it's created here https://github.com/credo-ai/credoai_lens/blob/develop/credoai/evaluators/privacy.py#L169 <- better, but requires us to go check every evaluator for this sort of behavior and to remember to do this for any future evaluators
3. Enable specifying evidence requirements in terms of the entire evidence object, not just the label. This would entail something like (on PP Editor) specifying `labels: [metric_type: ..., evaluator: ...], metadata: [subtype: ...]` <- this might be the best long-term solution since it provides a lot of flexibility for the policy pack creator to match arbitrary outputs from Lens (as long as the outputs from Lens are unique, which they presently are). Likely much more involved to implement.

Solution 2 is probably the best for right now. Solution 3 should be on the radar as a future feature.

esherman-credo:
## Describe your changes
Modified DataFairness and Privacy evaluators to create results as several Metric containers (one for each row in pre-results DataFrame). This allows passing subtype information as an additional label to the Container constructor (and subsequently the Evidence constructor).

Opening the PR now, as this fixes the two evaluators where this is known to be an issue. ~Going to look at other evaluators to see whether or not this issue is restricted to DataFairness + Privacy.~

UPDATE: This or a similar issue also applies to ~IdentityVerification, and RankingFairness.~ It is possible (have not confirmed, but I think it's unlikely) that it also applies to ~Equity and~ SurvivalFairness.

Security has a `subtype` thing going on, but the metrics are uniquely identified by the `metric_type` field...there are only 2 metrics.
Equity is fine.
IdentityVerification is confirmed to be fine. All metrics/tables created by the evaluator lead to unique labels without further mangling.
RankingFairness seems fine for now.

Final update: no existing test cases for Survival. I'm not going to test this right now to see what the label outputs look like
Other evaluators named above have been run & checked to determine what they're labeling scheme looks like.

Fixes https://github.com/credo-ai/credoai_lens/issues/265

See above comment about other evaluators. Reviewer can refuse to approve until confirmation that no other evaluators are subject to this issue.

N/A
N/A
No

## Extra-mile Checklist
- [x] I have thought expansively about edge cases and written tests for them
No new tests. This bug was identified as part of implementation of broader Platform+Lens+Connect integration test, which will catch this sort of bug once online.


IanAtCredo:
I don't have context on what this project is about. I don't think one-off fixes of these evaluators is the way to do this though.

Let's plan for next week to be bringing these "less used" evaluators up-to-date and make sure they are not breaking.

esherman-credo:
These are bugs encountered while trying to get evidence to properly push to Platform for the integration test.

esherman-credo:
Reopening for review in case we want to go this route to fix the bug. Can continue to discuss.

Only the two modified evaluators seem to have this subtyping issue. All other evaluators have been checked, except for SurvivalFairness.

PR #270, opened 2022-12-01
Feat/refactor tests

fabrizio-credo:
## Refactor integration tests

- Moved from class to single function test framework
- Split file into consistent init and testing part
- Added evaluator parametrization when possible

### Hotfix
- Added name and tags to all dummy models, required for governance part to run

### DIFF
--- a/credoai/artifacts/model/classification_model.py
+++ b/credoai/artifacts/model/classification_model.py
@@ -57,9 +57,11 @@ class DummyClassifier:
         Array containing the output of a model's "predict_proba" method
     """
 
-    def __init__(self, predict_output=None, predict_proba_output=None):
+    def __init__(self, name, predict_output=None, predict_proba_output=None, tags=None):
         self.predict_output = predict_output
         self.predict_proba_output = predict_proba_output
+        self.name = name
+        self.tags = tags
 
     def predict(self, X=None):
         return self.predict_output
--- a/tests/test_lens.py
+++ b/tests/test_lens.py
@@ -226,49 +226,24 @@ def test_privacy(init_lens_credit):
     pytest.assume(not gov._file_export(temp_file))
 
 
-def test_data_fairness(init_lens_classification):
+@pytest.mark.parametrize(
+    "evaluator",
+    [DataFairness, DataProfiler, ModelEquity, DataEquity, Security, Deepchecks],
+    ids=[
+        "DataFairness",
+        "DataProfiler",
+        "ModelEquity",
+        "DataEquity",
+        "Security",
+        "Deepchecks",
+    ],
+)
+def test_generic_evaluator(init_lens_classification, evaluator):
+    """
+    Any evaluator not requiring specific treatment can be tested here
+    """
     lens, temp_file, gov = init_lens_classification
-    lens.add(DataFairness())
-    lens.run()
-    pytest.assume(lens.get_results())
-    pytest.assume(lens.get_evidence())
-    pytest.assume(lens.send_to_governance())
-    pytest.assume(not gov._file_export(temp_file))
-
-
-def test_data_profiler(init_lens_classification):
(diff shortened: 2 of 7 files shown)


tests/test_lens.py
```diff
@@ -28,7 +29,9 @@
 
 from connect.governance import Governance
 
-import tempfile
+##################################################
+#################### Init ########################
+##################################################
 
 TEST_METRICS = [
```
esherman-credo, line 36: Can we move this to another file since we'll have other parametrizations in the near future?

tests/test_lens.py
```diff
@@ -28,7 +29,9 @@
 
 from connect.governance import Governance
 
-import tempfile
+##################################################
```
esherman-credo, line 32: Can you add to documentation (probably just in this file or in a tests readme) instructions on what components tests need?
Looks like: init function, test function.
How do we know what model + data to pass to the init function?

esherman-credo (changes requested):

credoai/artifacts/model/classification_model.py
```diff
@@ -57,9 +57,11 @@ class DummyClassifier:
         Array containing the output of a model's "predict_proba" method
     """
 
-    def __init__(self, predict_output=None, predict_proba_output=None):
+    def __init__(self, name, predict_output=None, predict_proba_output=None, tags=None):
```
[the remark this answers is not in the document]
fabrizio-credo: You mean add the typing? It is not checked in standard models.

tests/test_lens.py
```diff
@@ -46,324 +49,49 @@
...
-    gov = Governance()
-    my_pipeline = Lens(
-        model=classification_model,
-        assessment_data=classification_assessment_data,
-        training_data=classification_train_data,
-        governance=gov,
-    )
-    request.cls.pipeline = my_pipeline
+@pytest.fixture(scope="function")
+def temp_file(tmp_path):
```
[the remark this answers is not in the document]
fabrizio-credo: It's a pre existing fixture from pytest, it can be referred to without initializing it. It takes care of dumping everything after the test is ended. Similar to tempfile, but pytest native. See https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

tests/test_lens.py
fabrizio-credo: Yeah makes sense, I was debating whether to do it or not.

credoai/artifacts/model/classification_model.py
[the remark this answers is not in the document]
esherman-credo: It's enforced for classification model here: https://github.com/credo-ai/credoai_lens/blob/c52cf854acaf9bb610735506b319ae27cadebc10/credoai/artifacts/model/classification_model.py#L25

tests/test_lens.py
esherman-credo: 😲

tests/test_lens.py
esherman-credo: I think this part is still pending?

esherman-credo: @fabrizio-credo this sort of high-level description is a must.

It's not clear why we use the "credit" data for one test and the "classification" data for all the others. If someone wants to write a new test, they should be able to understand what exactly they need: which assume statements are the defaults to include, which data, under what circumstances does it make sense to modify conftest, etc.?

esherman-credo (changes requested):

tests/test_lens.py
fabrizio-credo: Added a readme file. Please consider this a prototype. I will expand on it when I finish the remaining work on tests during the week.

esherman-credo (approved):
lgtm

PR #277, opened 2022-12-05
update validation functions

IanAtCredo:
## Describe your changes
`TabularData` wasn't properly validating pre/post processing. This PR fixes that. 

Also added basic name validation.

* Should we just require the input to be a dataframe?

## Extra-mile Checklist

### DIFF
--- a/credoai/artifacts/data/tabular_data.py
+++ b/credoai/artifacts/data/tabular_data.py
@@ -5,7 +5,7 @@ from typing import Union
 import numpy as np
 import pandas as pd
 
-from credoai.utils.common import ValidationError
+from credoai.utils.common import ValidationError, check_array_like
 
 from .base_data import Data
 
@@ -66,24 +66,21 @@ class TabularData(Data):
 
         If y is convertible, convert y to pandas object with X's index
         """
-        if isinstance(y, (np.ndarray, list)):
-            pd_type = pd.Series
-            if isinstance(y, np.ndarray) and y.ndim == 2 and y.shape[1] > 1:
-                pd_type = pd.DataFrame
-            if self.X is not None:
-                y = pd_type(y, index=self.X.index, name="target")
-            else:
-                y = pd_type(y, name="target")
+        pd_type = pd.Series
+        if isinstance(y, np.ndarray) and y.ndim == 2 and y.shape[1] > 1:
+            pd_type = pd.DataFrame
+        if self.X is not None:
+            y = pd_type(y, index=self.X.index, name="target")
+        else:
+            y = pd_type(y, name="target")
         return y
 
     def _validate_X(self):
-        try:
-            pd.DataFrame(self.X)
-        except:
-            raise ValidationError("X must be able to be coerced into a dataframe")
+        check_array_like(self.X)
 
     def _validate_y(self):
         """Validation of Y inputs"""
+        check_array_like(self.y)
         if self.X is not None and (len(self.X) != len(self.y)):
             raise ValidationError(
                 "X and y are not the same length. "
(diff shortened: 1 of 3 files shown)


credoai/artifacts/data/tabular_data.py
```diff
@@ -52,16 +52,20 @@ def copy(self):
...
 
     def _process_y(self, y):
-        """Standardize y data"""
-        # if X is pandas object, and y is convertible, convert y to
-        # pandas object with X's index
+        """Standardize y data
+
+        If y is convertible, convert y to pandas object with X's index
+        """
         if isinstance(y, (np.ndarray, list)):
```
esherman-credo, line 69: What happens if this `isinstance()` call returns False? That would be the case if the user passed something entirely ill-formed --> seems like right now we're just assuming Y matches the docstring and is thus always a DF/Series, nd.Array, or List?

credoai/artifacts/data/tabular_data.py
```diff
@@ -72,25 +76,30 @@ def _process_y(self, y):
                 y = pd_type(y, name="target")
         return y
 
+    def _validate_X(self):
+        try:
+            pd.DataFrame(self.X)
```
esherman-credo, line 81: Technically someone could pass a dictionary and this wouldn't fail. They'd be violating the documentation but I _think_ it'd still end up in the right format based on other validation. Nevertheless, it might be better to explicitly forbid it through `isinstance` checking since there could be unanticipated issues with dictionaries.

Example:
```
>>> blarg = {'col1': [4, 5, 6], 
... 'col2': [1, 2, 3]}
>>> import pandas as pd
>>> pd.DataFrame(blarg)
   col1  col2
0     4     1
1     5     2
2     6     3
```

credoai/artifacts/data/tabular_data.py
```diff
@@ -72,25 +76,30 @@ def _process_y(self, y):
...
+            raise ValidationError(
+                "X and y are not the same length. "
+                + f"X Length: {len(self.X)}, y Length: {len(self.y)}"
+            )
+
+    def _validate_processed_X(self):
+        """Validate processed X"""
+        if len(self.X.columns) != len(set(self.X.columns)):
+            raise ValidationError("X contains duplicate column names")
+        if len(self.X.index) != len(set(self.X.index)):
```
esherman-credo, line 97: It would make sense to also check that `sensitive_feature.columns` doesn't intersect with `X.columns` if both objects are specified.

esherman-credo (changes requested):
Some extra validation is needed for y.

Other comments are nice-to-haves.

credoai/artifacts/data/tabular_data.py
IanAtCredo: I don't think that's something we want to forbid. The sensitive features are used for sensitive feature things - they could exist in X as well (to be passed to the model)

credoai/artifacts/data/tabular_data.py
IanAtCredo: I looked for validation of "array-like" as requested by sklearn. It doesn't seem like there is such validation. This seems like quite an odd case to be passed and not revealed by other validations

IanAtCredo:
@esherman-credo , found a sklearn util called `check_array` which seems almost like what we want. Added onto it to address your comments.

esherman-credo (commented):
Changes look good. I think we need to think harder about what to do about overlap between sensitive features and X.

credoai/artifacts/data/tabular_data.py
IanAtCredo: It won't do deceptively well... it will do really well and that is what we want. The features used for `sensitive_features` are distinct from the features used in X, which is the predictor features.

esherman-credo (approved):

PR #284, opened 2022-12-08
Add test-reports workflow for develop and main with badges

credo-nate:
## Describe your changes

test-reports.yml workflow will run on pushes to develop and main.  It can also be triggered for other branches but the badges will not be updated.  Badges are updated by pushing a generated badge to S3 that is referenced in the README.md.  The README.md references the badges for the main branch but the URL to retrieve the badges for develop can be discovered easily.

The permissions to write to the S3 bucket are limited to this workflow running on the develop and main branch.

Examples of the badges generated for the last run on this branch before lockdown in AWS:
Tests
[Coverage](https://credoai-cicd-public-artifacts.s3.us-west-2.amazonaws.com/credoai_lens/feature-add-badges/html/index.html)

This was a request in Slack.

- [x] I have built basic tests for new functionality (particularly new evaluators)

## Extra-mile Checklist


esherman-credo:
The badges don't seem to render for me on the front page of the branch:
image
Is that simply because the branch has been [removed from the workflow](https://github.com/credo-ai/credoai_lens/pull/284/files#diff-02a00b3701184653ca1230ead969855c3d5c7ec3472256d72d34cc6c9cbc3619R8) file?

.github/workflows/test-reports.yml
```diff
@@ -0,0 +1,81 @@
+name: Test Reports
+
+on:
+  workflow_dispatch:
+  push:
```
esherman-credo, line 5: Just so I understand: normal test suite will run on all pushes to PRs. The coverage suite will only run when we merge (we're never pushing directly to develop that I'm aware of)?

credo-nate:
> The badges don't seem to render for me on the front page of the branch: image Is that simply because the branch has been [removed from the workflow](https://github.com/credo-ai/credoai_lens/pull/284/files#diff-02a00b3701184653ca1230ead969855c3d5c7ec3472256d72d34cc6c9cbc3619R8) file?

Yes.  I updated the README.md to reference the main branch as that is what it would be when we merge to main.  The goal of this is to show the status of the workflow that is run against the stable/release branch which is what main is right now.  I could entertain the idea that develop is more important to show at this time as it represents the next release.  Unfortunately, markdown doesn't have variable substitution that would allow us to make it dynamic per branch.  As I write that, I wonder if I can do something on the badge hosting side to server based on CORS...

esherman-credo:
Sounds good, just wanted to confirm

.github/workflows/test.yml.orig
```diff
@@ -0,0 +1,45 @@
+name: Test
```
esherman-credo, line 1: This file is just for backup?

esherman-credo (approved):
Lgtm

.github/workflows/test.yml.orig
credo-nate: removed.  good catch.

.github/workflows/test-reports.yml
credo-nate: The PR workflow will run on all PRs.  This version of the PR workflow will run on push to develop and main which is effectively the merge to those respective branches.

PR #290, opened 2022-12-21
Feat/image data stable

esherman-credo:
## Describe your changes
This is a feature branch implementing support for tensor-based data and neural network models.
For Credo-internal developers, the proposed changes are detailed [here](https://credo-ai.atlassian.net/wiki/spaces/MVP/pages/551845889/Lens+Design+Doc+Image+Data+Support).

Summary of Changes (in progress, more to come):
- BaseModel 
    - Group init functionality in `process_model` function
    - Split validation functionality into 2 functions:
        1. Checks that required callable functions are present (existing functionality)
        2. Checks that various details of the framework match up with the model wrapper type; throws warnings rather than errors
- ClassificationModel
    - Add support for Keras-style classifiers (highly restrictive, w/ warnings if assumptions are not met)
        - Assume Sequential-style model w/ Dense last (or 2nd to last) layer + (softmax or sigmoid)
        - Implement `__post_init__()` functionality for keras predictions
            - Depends on sigmoid (label outputs; no `predict_proba`) vs. softmax (probability outputs; `predict_proba` by default -> can infer `predict`)
-DummyClassifier
    - Passing `X` + `y` artifacts to data wrapper when they're already wrapped
    - If `X` and `y` are separate in the user's non-Lens workflow, passing these to Lens will be done in the same way as for `TabularData`
    - If `X` and `y` are wrapped in one object (such as a `tf.data` or `keras.utils.Sequence` object) then passing them to the Lens wrapper requires separating them
    - **Change: Add ability to pass `model_like` to `DummyClassifier`**
        - This does not fully address usability issue above (e.g., if user wants to run dataset assessments on `X` and `y`)
        - Allows user to avoid excessive computation by Lens, while still retaining model details for, e.g., the ModelProfiler evaluator
- Base `Data`
    - Converted `process` functions to non-abstract (committed)
        - Can support tensor data (from Keras + TF, at least; see supported inputs to predict [here](https://www.tensorflow.org/api_docs/python/tf/keras/Model#predict)) with base `Data` class
        - Working example of data-on-disk model fitting + Lens evaluation exists (not pushed to GH)
<ins>Evaluator Validation</ins>
- Security evaluator needed to be refactored (a bit) to use ART's `TensorflowV2Classifier` rather than their `KerasClassifier`
    - The latter does not support eager execution
    - The former is ART's current workaround -> eager execution is important for overall TF/Keras support (allows running things on-demand rather than as a batched graph execution)
    - Tests passing

<ins>Lens Validation</ins>
Validation of Model + Data --> At Lens init stage, we now verify that `predict`, `predict_proba`, and `compare` (whichever are relevant to the provided model) works for the provided data. Throws an error and prevents instantiating/running evaluators without first checking Model + Data compatibility.

<ins>Evaluator Validation</ins>
Established a starting point for streamlining/unifying artifact checking. Converted `check_artifact_for_nulls` to `check_data...` (name reflects what it's doing) and added options to only check some parts of the artifact (i.e., a subset of `X`, `y`, and `sensitive_features`) for nulls rather than checking all parts. Functionality doesn't fundamentally change but makes requirements more explicit from function arguments: `check_X`, `check_y`, and `check_sens` (all boolean).

- If datatype is completely arbitrary or generator-like, we have no way of checking; need the user to do this before wrapping
- This revised null-checker expands capabilities to several non-Pandas types
  
https://credo-ai.atlassian.net/browse/DSP-344

- Need to fix/merge the Keras support in the ModelProfiler
- Tests! Tests! Tests!
- Confirm documentation is properly built

- [x] If new libraries have been added, I have checked that [readthedocs](https://readthedocs.org/projects/credoai-lens/) API documentation is constructed correctly
- [x] Will this be part of a major product update? If yes, please write one phrase about this update.
    - See above. This expands Lens functionality to support Keras models (and, by default, some TF models). This remains relatively experimental -> the space of models and data that "work" is likely much larger than what we have explicitly developed for. We have tried to implement warnings where possible. Some growing pains are likely.

## Extra-mile Checklist

[DSP-344]: https://credo-ai.atlassian.net/browse/DSP-344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


esherman-credo:
[Linking the per-evaluator validation requirements](https://docs.google.com/spreadsheets/d/1tI70VHhF_9bqHGtEkcBMxR82WaDMYoQVjBkeGxugy_c/edit#gid=0). I'm shooting to have this implemented such that one can trivially read an evaluator's assumptions/requirements from the `validate_artifacts` function. E.g. in below screenshot, we can see that Performance.py has the requirements:

1. The metrics, X, and y supplied to the evaluator all need to be non-null (as objects)
2. X and y also need to not contain object (i.e. the internals)

Moreover, we can see that sensitive features doesn't need to be checked for nullness, which is a hint that that sub-artifact isn't used in this evaluator (not otherwise obvious without looking carefully through the code or making an assumption based on the evalutor's docstring).

image

credoai/lens/__init__.py
```diff
@@ -6,3 +6,4 @@
 
 from .lens import Lens
 from .pipeline_creator import PipelineCreator
+from .lens_validation import *
```
fabrizio-credo, line 9: Is it necessary to make `lens_validation` functinality avaiable at this level since it is only used by `Lens`? `import *` is generally not recommended.

credoai/evaluators/fairness.py
```diff
@@ -59,9 +53,18 @@ def __init__(
...
+        if not check_pandas(self.data.sensitive_features):
+            raise ValidationError(
+                "Sensitive features must be a Pandas DataFrame or Series"
+            )
+        check_data_for_nulls(
+            self.data, "Data", check_X=True, check_y=True, check_sens=True
+        )
+        # check_data_instance(self.data, TabularData)
+        # check_existence(self.data.sensitive_features, "sensitive_features")
+        # check_data_for_nulls(self.data, "Data")
```
fabrizio-credo, line 67: We can remove the extra lines

credoai/evaluators/model_profiler.py
```diff
@@ -77,15 +81,24 @@ def __init__(self, model_info: Optional[dict] = None):
...
     def _validate_arguments(self):
         check_existence(self.model, "model")
 
     def evaluate(self):
         basic = self._get_basic_info()
         res = self._get_model_params()
+        if "keras" in str(self.model_type):
+            # Hack for keras demo, TODO: remove when merging to dev
+            self.results = [
+                TableContainer(
```
fabrizio-credo, line 96: TableContainer is not imported.

fabrizio-credo (commented):
Amazing work, I will leave approval to @IanAtCredo. I have just general comments and spotted a missing import which would be evident once new tests are added.

Something else I spotted, some evaluators (e.g., Privacy) have `check_model_instance` as part of their validation, but in general the `DummyModel` is not included in allowed models. That would lead to failure if the user attempts to use a dummy model.

For example:
`check_model_instance(self.model, ComparisonModel)` should be:
`check_model_instance(self.model, (ComparisonModel, DummyComparisonModel))`

This has been here since time immemorial, could you address it?

credoai/evaluators/utils/validation.py
```diff
@@ -50,6 +59,48 @@ def check_existence(obj, name=None):
...
+        global_logger.warning(message)
+    return nulls
+
+
+#################################
+# Checking evaluator requirements
+#################################
+
+
+def check_data_for_nulls(obj, name, check_X=True, check_y=True, check_sens=True):
```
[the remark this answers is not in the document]
esherman-credo: Not necessarily. This could refer to a Tensor or other object that doesn't have a `name` attribute. The `name` variable is only used for error outputting to say "This thing you provided (which you called `name`) contains nulls when it shouldn't"

credoai/evaluators/fairness.py
```diff
@@ -59,9 +53,15 @@ def __init__(
 
     def _validate_arguments(self):
         check_existence(self.metrics, "metrics")
-        check_data_instance(self.data, TabularData)
-        check_existence(self.data.sensitive_features, "sensitive_features")
-        check_artifact_for_nulls(self.data, "Data")
+        check_existence(self.data.X, "X")
+        check_existence(self.data.y, "y")
+        if not check_pandas(self.data.sensitive_features):
```
IanAtCredo, line 58: is this redundancy? Sensitive features must always be pandas objects.

credoai/evaluators/model_profiler.py
```diff
@@ -77,15 +82,23 @@ def __init__(self, model_info: Optional[dict] = None):
...
+        self.model_internal = self.model.model_like
+        self.model_type = type(self.model_internal)
 
     def _validate_arguments(self):
         check_existence(self.model, "model")
 
     def evaluate(self):
         basic = self._get_basic_info()
         res = self._get_model_params()
+        if "keras" in str(self.model_type):
```
IanAtCredo, line 94: why this check rather than checking framework in model_info?

credoai/evaluators/utils/validation.py
```diff
@@ -29,19 +38,19 @@ def check_model_instance(obj, inst_type, name="Model"):
...
             message = f"Feature {feature_name} not found in dataframe {name}"
             raise ValidationError(message)
-    if isinstance(df, Series):
+    if isinstance(df, pd.Series):
         if not df.name == feature_name:
             message = f"Feature {feature_name} not found in series {name}"
             raise ValidationError(message)
 
 
 def check_existence(obj, name=None):
```
IanAtCredo, line 51: I'm confused about what this function is for. Why the pandas check? (I know you didn't write it though)

credoai/lens/lens_validation.py
```diff
@@ -0,0 +1,172 @@
...
+                mini_pred.shape != data.y[:batch_size].shape
+            ):
+                raise Exception("Predictions have mismatched shape from provided y")
+            if isinstance(data.y, pd.Series) and (
+                mini_pred.shape != data.y.head(batch_size).shape
+            ):
+                raise Exception("Predictions have mismatched shape from provided y")
+        except Exception as e:
+            raise ValidationError(
+                "Lens.model predictions do not match expected form implied by provided labels y.",
```
IanAtCredo, line 52: Isn't this error too specific? If the predict function flat out fails, don't we want to say that and reproduce the error? I'm worried we are catching an exception that we want to raise.

credoai/lens/lens_validation.py
```diff
@@ -0,0 +1,172 @@
...
+                    raise Exception(
+                        "`predict_proba` outputs invalid. Binary outputs should be <= 1."
+                    )
+        except Exception as e:
+            raise ValidationError(
+                "Lens.model outputs do not match expected form implied by provided labels y.",
+                e,
+            )
+
+    if "compare" in model.__dict__.keys():
```
IanAtCredo, line 80: Do you think these validations could/should be defined within the model? It seems they are mostly dependent on the model functionality, with the data as input. What do you think of each model having a `validate_against_data` function that this lens_validation function calls?

IanAtCredo (changes requested):
Looks awesome. I have a few small comments.

credoai/evaluators/fairness.py
esherman-credo: Yes it is. I can remove it. Is the fact that it makes validation steps directly readable off of the evaluator's validation function a good enough reason to keep it?

credoai/artifacts/model/classification_model.py
```diff
@@ -16,23 +27,47 @@ class ClassificationModel(Model):
...
         A binary or multi-class classification model or pipeline. It must have a
-            `predict` function that returns array containing the class labels for each sample.
+            `predict` function that returns an array containing model outputs for each sample.
             It can also optionally have a `predict_proba` function that returns array containing
-            the class labels probabilities for each sample.
+            the class label probabilities for each sample.
+            If the supplied model_like is from the sklearn or xgboost framework, `predict` is assumed
+            to return a column vector with a single value for each sample (i.e. thresholded predictions).
+            If the supplied model_like is from the Keras framework, the assumed form of `predict` outputs
+            depends on the final-layer activation. If softmax, wrapper assumes return is a matrix with
```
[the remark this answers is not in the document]
esherman-credo: Sigmoid for binary classification returns just a single value per sample. If someone's architecture is set up to return a vector, that corresponds to multi-output classification (e.g. probabilities of being in each of several non-mutually-exclusive classes). In particular, the probabilities need not sum to 1.
Softmax, the probabilities do sum to 1 and it generally corresponds to single-output multi-class. I'm not totally sure how you'd do multi-output classification using softmax...maybe multiple heads or something.

We can do some extra work to enable multi-output classification (via sigmoid) but that seems like a next step.
For this PR, I assume softmax will always be a vector (length >= 2) per sample and treat it as predict_proba. 
If they are doing sigmoid, I had written the code in a way that assumes they take the extra step of thresholding/rounding to output discrete labels.

On further thought, that assumption probably doesn't make sense: better to assume probability outputs (which we can round, not argmax) than to assume discrete predictions -> can always discretize from cts. Cannot get cts. from discrete.

credoai/artifacts/model/classification_model.py
```diff
@@ -16,23 +27,47 @@ class ClassificationModel(Model):
...
+        Additional metadata to add to model
+        E.g., {'model_type': 'binary_classification'}
     """
 
     def __init__(self, name: str, model_like=None, tags=None):
         super().__init__(
             "CLASSIFICATION",
             ["predict", "predict_proba"],
             ["predict"],
+            # TODO this will not work once we incorporate PyTorch
```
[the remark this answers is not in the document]
esherman-credo: I think the latter. In my limited experience, PyTorchers just use `forward` (or `__call__`) everywhere. A quick google for "pytorch predict function" turned up various stack overflow discussions about how to get _predictions_ but generally none that explicitly implement a `predict` function.

I think we should leave that refactor for another day but open to doing it now (it would end up all of our existing model wrappers).

credoai/evaluators/utils/validation.py
esherman-credo: I actually have no clue. It may be that this is supposed to check for emptiness?

If so, inside the pandas if-statement, we can use the `.empty` attribute.
The other if-statement is assuming any other provided object follows the semantics of built-in python data types where emptiness can be checked with a simple `not`. Not sure if this is liable to fail for any non-standard library data types we use --> should be fine for Tensors, at least, since these need to be initialized with some value (this is driven by the fact that tensorflow is backed by C for speed and so the tensors are immutable once instantiated).

credoai/lens/lens_validation.py
esherman-credo: The call to the predict function happens in the helper function on line 38. There's no specific error message in there so if the predict function fails, then we'll catch and raise the message on line 52 + the original exception (note the addition of the exception with `, e)` on line 53/54.

The checks on lines 39-49 kick in only if `predict` itself works but there's somehow a mismatch between `y_pred` and the supplied `y_true`.

credoai/lens/lens_validation.py
esherman-credo: I don't really see it as being obviously better one way or the other.

This model + data validation step is only going to run once: when model + data are passed to the Lens object. I think there are equally strong arguments for that validation being a feature of the model (it's about its behavior on data) vs. Lens (we care about whether the things you passed for evaluation work together as a unit).

I tried to think of which way would lead to easier development down the road and I think the answer there is also "it's about equal". 
- In the paradigm where validation is a feature of the model, we'd make this existing code more modular (cut down on the if-statements in `check_model_data_consistency()`) and only call these helpers more places. We'd add helpers when there's new functionality like `my_funky_predict_function()`. If there's a new data type we'd have to go through all the existing models (or merely the helpers) to add some if statements (see e.g., lines 118-124 in the current version of this file).
- In the paradigm where validation lives primarily in lens (status quo in this PR), future development would mean adding ugly if-statements (that I think we need in scenario 1 anyway) to `check_model_data_consistency` and crafting a new helper for whatever functionality (for new models/prediction functions) or adding if-statements to helpers for new data types

credoai/artifacts/model/classification_model.py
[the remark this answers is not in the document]
IanAtCredo: Created an issue.

credoai/evaluators/fairness.py
IanAtCredo: I removed it. I don't think we want redundant validation.

credoai/evaluators/utils/validation.py
IanAtCredo: Updated based on your suggestion.

credoai/artifacts/model/classification_model.py
```diff
@@ -81,42 +88,50 @@ def __post_init__(self):
...
-                    self.__dict__["predict_proba"] = lambda x: pred_func(x)[:, 1]
-                elif (
-                    len(self.model_like.layers[-1].output_shape) == 2
-                    and self.model_like.layers[-1].output_shape[1] > 2
-                ):
-                    self.__dict__["predict_proba"] = pred_func
-                else:
-                    pass
-                    # predict_proba is not valid (for now)
+            if self.model_like.layers[-1].output_shape == (None, 1):
```
IanAtCredo, line 91: Why was the check for `pred_func` not needed?

IanAtCredo (approved):
looks good. One question about pred_func

credoai/artifacts/model/classification_model.py
esherman-credo: It is. Added back.

PR #292, opened 2022-12-22
Feat/data fairness simplification

fabrizio-credo:
## Describe your changes
Extracted run_cv and get demo_parity from data fairness

## Extra-mile Checklist

### DIFF
--- a/credoai/evaluators/data_fairness.py
+++ b/credoai/evaluators/data_fairness.py
@@ -238,38 +238,16 @@ class DataFairness(Evaluator):
         for col in self.categorical_features_keys:
             self.X[col] = self.X[col].astype("category").cat.codes
 
-        discrete_features = [
+        self.discrete_features = [
             True if col in self.categorical_features_keys else False
             for col in self.X.columns
         ]
 
         # Use the right mutual information methods based on the feature type of the sensitive attribute
         if is_categorical(self.sensitive_features):
-            sensitive_feature = self.sensitive_features.cat.codes
-            mi = mutual_info_classif(
-                self.X,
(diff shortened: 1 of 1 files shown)


IanAtCredo (changes requested):
These seems like a WIP.

The goal of the helper functions is so we can look through them (or metrics) and understand what the evaluator does. So things like `_make_pipe` shouldn't be there and aspects of the other gigantic functions should be better modularized.

I moved `_make_pipe` into a FeatureInferenceAttacker class (I liked that description of it). But other functionality still seems needlessly complicated. `assess_balance_metrics` is a prime offender.

Did you not included functions like `calculate_mutual_information` because it seemed dependent on too many underlying arguments?

fabrizio-credo:
In general I was thinking about what functions to extract in terms of topic/re-usability. The inferenceattack one seemed to be a good candidate for an external method used by data fairness. The other methods seemed more connected to the class function itself rather than stuff to be re-used elsewhere, therefore I left it in. Agreed that there can be more simplification.

IanAtCredo:
> In general I was thinking about what functions to extract in terms of topic/re-usability. The inferenceattack one seemed to be a good candidate for an external method used by data fairness. The other methods seemed more connected to the class function itself rather than stuff to be re-used elsewhere, therefore I left it in. Agreed that there can be more simplification.

I think there are a few things here. Reusability is one, for sure. But even if functionality is _only_ used by one evaluator, the fact that it is buried in the class makes it difficult to document. Let's briefly sync later today on what we want/need from further simplification (if any)

IanAtCredo (approved):
lg

PR #304, opened 2023-01-17
Only run validation if input data is actually provided

esherman-credo:
## Describe your changes
Fixes https://credo-ai.atlassian.net/jira/software/c/projects/DSP/boards/29?modal=detail&selectedIssue=DSP-395

## Extra-mile Checklist

### DIFF
--- a/credoai/lens/lens_validation.py
+++ b/credoai/lens/lens_validation.py
@@ -81,7 +81,7 @@ def check_model_data_consistency(model, data):
                 e,
             )
 
-    if "compare" in model.__dict__.keys() and data.comps is not None:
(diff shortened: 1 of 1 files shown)


amrasekh (approved):
For my candidate scoring use case, it woks as expected.

IanAtCredo (approved):
Looks good, but one thought - should we include the check for data.y and data.X in the `check_prediciton_model_output` function itself?

Basically, if the data passed doesn't have the correct information (or is None), the function just returns the info rom the model predict?

Happy with this implementation too.

esherman-credo:
We aren't using y_true for anything except to compare against the outputs. It might make sense to move the check for data.X but then we'd be separating them and that feels like an overcomplication.

PR #305, opened 2023-01-17
Fixing validation for Dummy models which don't specify a prediction function

esherman-credo:
## Describe your changes
For DummyClassifier: change prediction functions to be lambdas so that the functions don't refer to attributes that don't exist if the user has not provided one of predict/predict_proba

For DummyClassifier & DummyRegressor: change validation to compare ALL outputs of predict/predict proba to y. Other models jsut predict on a sample for latency reasons...Better to check full output for dummys

https://credoai.slack.com/archives/C04612H3P7T/p1673996354009329

## Extra-mile Checklist

### DIFF
--- a/credoai/artifacts/model/classification_model.py
+++ b/credoai/artifacts/model/classification_model.py
@@ -167,15 +167,19 @@ class DummyClassifier:
             self.predict_output = check_array(
                 predict_output, ensure_2d=False, allow_nd=True
             )
+            self.__dict__["predict"] = lambda x=None: self.predict_output
+
         if predict_proba_output is not None:
             self.predict_proba_output = check_array(
                 predict_proba_output, ensure_2d=False, allow_nd=True
             )
+            self.__dict__["predict_proba"] = lambda x=None: self.predict_proba_output
+
         self.name = name
         self.tags = tags
 
-    def predict(self, X=None):
-        return self.predict_output
+    # def predict(self, X=None):
+    #     return self.predict_output
 
-    def predict_proba(self, X=None):
-        return self.predict_proba_output
+    # def predict_proba(self, X=None):
+    #     return self.predict_proba_output
--- a/credoai/artifacts/model/classification_model.py
+++ b/credoai/artifacts/model/classification_model.py
@@ -163,23 +163,17 @@ class DummyClassifier:
         tags=None,
     ):
         self.model_like = model_like
-        if predict_output is not None:
-            self.predict_output = check_array(
-                predict_output, ensure_2d=False, allow_nd=True
-            )
-            self.__dict__["predict"] = lambda x=None: self.predict_output
-
-        if predict_proba_output is not None:
-            self.predict_proba_output = check_array(
-                predict_proba_output, ensure_2d=False, allow_nd=True
-            )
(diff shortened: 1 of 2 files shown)


credoai/artifacts/model/classification_model.py
```diff
@@ -167,15 +167,19 @@ def __init__(
...
         if predict_proba_output is not None:
             self.predict_proba_output = check_array(
                 predict_proba_output, ensure_2d=False, allow_nd=True
             )
+            self.__dict__["predict_proba"] = lambda x=None: self.predict_proba_output
+
         self.name = name
         self.tags = tags
 
-    def predict(self, X=None):
```
IanAtCredo, line 177: Replace with a helper functions like:
```
def _wrap_array(self,  array):
      return lambda x: array

def _build_functionality(self., function_name, array):
     if array is not None:
             array = check_array(array, ensure_2d=False, allow_nd=True)
             self.__dict__[function_name] = self._wrarp_array(array)


```

Add documentation, and use those functions to define

credoai/lens/lens_validation.py
```diff
@@ -34,52 +36,68 @@ def check_model_data_consistency(model, data):
...
-            if isinstance(data.y, pd.Series) and (
-                mini_pred.shape != data.y.head(batch_size).shape
-            ):
-                raise Exception("Predictions have mismatched shape from provided y")
-        except Exception as e:
-            raise ValidationError(
-                "Lens.model predictions do not match expected form implied by provided labels y.",
-                e,
-            )
+        if isinstance(model.model_like, (DummyClassifier, DummyRegression)):
```
IanAtCredo, line 39: Doesn't seem great to have validation dependent on Dummy class. Why can't we use the same functionality as the rest of the validation? `predict` and `predict_proba` should work identically to other models.

IanAtCredo (changes requested):
Added comments for changes.

credoai/lens/lens_validation.py
esherman-credo: Right now for normal models, I'm doing `predict` on some small chunk of `data.X` to avoid too much heavy computation. I then compare shapes/sizes of predictions and `data.y`

For `DummyClassifier.predict(X)`, you just get the whole prediction array, whereas `data.y` is still subsetted.

One solution would be to do predictions for all data (no batching). The problem in that case is that image data would take forever.
Another solution is to set `batch_size = len(data.y)` for Dummy models.

esherman-credo: Alright I reworked this a little bit. 

The only place batch was actually used (once set) was to compare overall shapes (n_samples, output1, output2, ...).

In the reworking, I check that the per-sample shape is the same (so check that the (output1, output2, ...) part is the same shape). We'll implicitly assume n_samples in predictions vs. `data.y` are the same here and rely on downstream library functions (e.g. `sklearn.accuracy_score`) to throw an error if the user provided mismatched sample sizes for `data.y` and DummyClassifier (note that it's not possible for this to be an issue with vanilla ClassificationModel because that uses `data.X` and `data.y` which have to be the same).

Maybe there's an even better solution that pushes this into the DummyClassifier??

PR #311, opened 2023-01-19
Multiple Validation Fixes for DummyClassifiers

esherman-credo:
## Describe your changes
Removed second passage of `message` in logger warning message.
Remove requirement that `data.X` be supplied for Performance and Fairness evaluators
Update `model.type` assignment for DummyClassifier and Keras classifiers.

Various slack messages...

## Extra-mile Checklist

### DIFF
--- a/credoai/artifacts/model/classification_model.py
+++ b/credoai/artifacts/model/classification_model.py
@@ -69,7 +69,7 @@ class ClassificationModel(Model):
         except:
             message = """Provided model is from unsupported framework. 
             Lens behavior has not been tested or assured with unsupported modeling frameworks."""
-            global_logger.warning(message, message)
+            global_logger.warning(message)
 
     def __post_init__(self):
         """Conditionally updates functionality based on framework"""
--- a/credoai/artifacts/model/classification_model.py
+++ b/credoai/artifacts/model/classification_model.py
@@ -92,9 +92,15 @@ class ClassificationModel(Model):
                 if self.model_like.layers[-1].output_shape == (None, 1):
                     # Assumes sigmoid -> probabilities need to be rounded
                     self.__dict__["predict"] = lambda x: pred_func(x).round()
+                    # Single-output sigmoid is binary by definition
+                    self.type = "BINARY_CLASSIFICATION"
                 else:
                     # Assumes softmax -> probabilities need to be argmaxed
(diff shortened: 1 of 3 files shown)


credoai/artifacts/model/classification_model.py
```diff
@@ -117,11 +123,16 @@ def __post_init__(self):
...
         elif self.model_info["framework"] == "credoai":
             # Functionality for DummyClassifier
-            self.model_like = getattr(self.model_like, "model_like", None)
+            if self.model_like.model_like is not None:
+                self.model_like = self.model_like.model_like
             # If the dummy model has a model_like specified, reassign
             # the classifier's model_like attribute to match the dummy's
             # so that downstream evaluators (ModelProfiler) can use it
 
+            self.type = self.model_like.type
```
IanAtCredo, line 132: Do we need to do it like this? Can't we just count the number of unique values that are created by the predict function?

Or perhaps we can do this to automatically determine the type within DummyClassifier...

credoai/artifacts/model/classification_model.py
```diff
@@ -141,6 +152,13 @@ class DummyClassifier:
     model_like : model_like, optional
         While predictions are pre-computed, the model object, itself, may be of use for
         some evaluations (e.g. ModelProfiler).
+    binary_clf : bool, optional, default = True
```
IanAtCredo, line 155: maybe an "auto" option would be useful? When "auto", classification will be determined by the `predict_output` vector

IanAtCredo (approved):
Non-blocking questions/suggestions for alternative implementations.

amrasekh (approved):

credoai/artifacts/model/classification_model.py
esherman-credo: That occurred to me but it's possible the supplied predictions don't actually contain samples with labels for all possible classes and thus we might incorrectly infer that it's supposed to be a binary problem when it's really multi-class.

credoai/artifacts/model/classification_model.py
esherman-credo: See above comment. Imagine `predict_output` is [0, 0, 0...] because the event is rare and the model can't tease out when something is likely to be class 1. Auto would treat this as...binary? What if there are two rare events (class 1 and 2)? We'd miss that.

PR #314, opened 2023-01-25
Fix/model validation

esherman-credo:
## Describe your changes
Model validation: add support for pd.Series as a datatype for X
Predict_proba functionality: for custom (non-Dummy) classifiers, we're going to allow users to "signal" to Lens that we should treat their `predict` and `predict_proba` functions as being sklearn-like. If the user adds a `frameworks_like` attribute (a list) to their custom classifier and pass that classifier as the `model_like` argument to `ClassificationModel`, we will check the contents of `framework_like`. If `sklearn` is in the attribute, we will take steps similar to existing `framework` checking and functionality updating. Note that _multiple_ frameworks can be specified. For instance, a user might want to pass a pytorch model that supports both `forward()` (default pytorch behavior) and custom-implemented `predict` and `predict_proba` behavior a la sklearn. In this case, the user would set an attribute in their custom model `frameworks_like = ['sklearn', 'pytorch']`. Full pytorch support not yet implemented.

Zoom discussion

Full support of `frameworks_like` not yet implemented.

## Extra-mile Checklist

### DIFF
--- a/credoai/lens/lens_validation.py
+++ b/credoai/lens/lens_validation.py
@@ -103,7 +103,7 @@ def check_model_data_consistency(model, data):
             )
 
 
-def check_prediction_model_output(fn, data, batch: int = 1):
+def check_prediction_model_output(fn, data, batch_in: int = 1):
     """
     Helper function for prediction-type models. For use with `check_model_data_consistency`.
 
@@ -122,16 +122,16 @@ def check_prediction_model_output(fn, data, batch: int = 1):
         since this could be large and computationally expensive.
     """
     mini_pred = None
-    batch_size = batch
+    batch_out = batch_in
     if isinstance(data.X, np.ndarray):
         mini_pred = fn(np.reshape(data.X[0], (1, -1)))
     elif isinstance(data.X, (pd.DataFrame, pd.Series)):
-        mini_pred = fn(data.X.head(batch))
+        mini_pred = fn(data.X.head(batch_in))
     elif isinstance(data.X, tf.Tensor):
         mini_pred = fn(data.X)
     elif isinstance(data.X, tf.data.Dataset) or inspect.isgeneratorfunction(data.X):
         one_batch = next(iter(data.X))
-        batch_size = len(one_batch)
+        batch_out = len(one_batch)
         if len(one_batch) >= 2:
             # batch is tuple
             # includes y and possibly weights; X is first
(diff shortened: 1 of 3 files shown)


fabrizio-credo (approved):

amrasekh:
@esherman-credo  I have [this](https://github.com/credo-ai/customer_demos/blob/prod/prod/d11_sentiment_analysis/temporary/loan_classification.ipynb) model that throws error on this branch. 
The model is essentially an `sklearn` classification model that is wrapped to function based on a custom threshold (a very common scenario)
Is this PR envisioned to accommodate for it?
There is also [this](https://github.com/credo-ai/customer_demos/blob/prod/prod/d11_sentiment_analysis/sentiment_analysis.ipynb) model which is a similar, but more complicated one, that is throwing the same error as well.
Feel free to run the notebooks for tests, etc. Thanks!

amrasekh (approved):
Tested with a number of sklearn-like custom models and all work as desired.

credoai/lens/lens_validation.py
```diff
@@ -125,8 +125,8 @@ def check_prediction_model_output(fn, data, batch: int = 1):
     batch_size = batch
```
fabrizio-credo, line 125: Why the reassigning of batch?

credoai/lens/lens_validation.py
```diff
@@ -125,8 +125,8 @@ def check_prediction_model_output(fn, data, batch: int = 1):
...
     if isinstance(data.X, np.ndarray):
         mini_pred = fn(np.reshape(data.X[0], (1, -1)))
-    elif isinstance(data.X, pd.DataFrame):
-        mini_pred = fn(data.X.head(1))
+    elif isinstance(data.X, (pd.DataFrame, pd.Series)):
+        mini_pred = fn(data.X.head(batch))
```
fabrizio-credo, line 129: Should it be batch_size since you reassigned above?

fabrizio-credo (approved):
A minor comment, it looks OK to me.

credoai/lens/lens_validation.py
esherman-credo: I wanted to avoid confusion between input `batch` and output `batch_size` since its value could be changed depending on the data type of `X`.

This is a primarily internal function so I'm not sure it matters. I can change to batch_out and batch_in.

PR #316, opened 2023-02-02
Feat/striate installation

IanAtCredo:
## Describe your changes
Added the ability to install core functionality separate from heavier functionality. Please test various installations and missing requirements!

## Extra-mile Checklist


credo-nate (approved):
`test.yml` looks good.

requirements.txt
```diff
@@ -10,31 +10,13 @@ python-dotenv>=0.19.2
...
 # required for ShapExplainer evaluator
 shap>=0.41.0
 
-# required for Deepchecks evaluator
-deepchecks>=0.9.1
-
 # required for RankingFairness evaluator
 finsfairauditing>=0.0.2
 
-# required for NLP evaluator
```
esherman-credo, line 35: These are fully cut? Seems like they're not in the extras file

credoai/evaluators/utils/utils.py
```diff
@@ -15,3 +19,34 @@ def all_subclasses(cls):
     return set(cls.__subclasses__()).union(
         [s for c in cls.__subclasses__() for s in all_subclasses(c)]
     )
+
+
+def list_evaluators_exhaustive():
+    """List all defined evaluation modules"""
+    evaluator_path = os.path.dirname(os.path.dirname(__file__))
```
esherman-credo, line 26: Will we ever have a case where evaluators are organized into further subfolders?

If so, this would need to be changed (heads up, but don't think it's a present problem to worry about)

setup.py
```diff
@@ -31,8 +34,14 @@
     doc_requirements = [line.strip() for line in f]
 
 dev_requirements += doc_requirements
+full_requirements = extras_requirements
```
esherman-credo, line 37: So my understanding is full = base + extra, all = base + extra + dev. The naming is a bit confusing since "full" and "all" could both be seen as meaning everything.

Maybe "docs", "extended", and "full"? Others for "full" could be "beta" or "experimental".

esherman-credo (changes requested):
The files look good on review.
I am getting the following error on a cursory import of Lens.
Environment is fresh install with requirements.txt (the "base" version of Lens) --> seems we'll need to refactor the Keras validation checking in `model_utils.py` to be elsewhere/only get imported if TF/Lens-full is installed (those functions won't get called unless someone is using a Keras model so they'd have the necessary installations anyway)


image

requirements.txt
IanAtCredo: Yes. Until we bring the NLP evaluator back.

credoai/evaluators/utils/utils.py
IanAtCredo: I think this actually goes into subfolders, but yeah, this is based on our current organization.

setup.py
IanAtCredo: I just removed the "all". You can install both dev and extras using credoai_lens[full,dev]

amrasekh (approved):
Reviewed the changes and tested installation of core on fresh kernel and all worked as expected.

credoai/utils/model_utils.py
```diff
@@ -1,11 +1,15 @@
 import warnings
-from credoai.utils import global_logger
 
 from sklearn.base import is_classifier, is_regressor
 from sklearn.ensemble import RandomForestClassifier
 from sklearn.utils import multiclass
 
-from tensorflow.keras import layers
+from credoai.utils import global_logger
```
esherman-credo, line 7: I don't think this file needs the "if tf_exists" paradigm in the `validate_keras_clf` function since we'd only be calling that function if keras is successfully imported elsewhere. That said, it might be good to put it in anyway.

esherman-credo (approved):

PR #318, opened 2023-03-08
WIP Dsp 143 nlp evaluator refactor

IanAtCredoAI:
## Describe your changes
This PR brings back in the NLP evaluator, this time described as an LLM_scorer. It has three main ways of scoring systems:

1. Use anthropic [evals](https://github.com/anthropics/evals). These are datasets that specify a dimension as well as a number of yes/no prompts indicating adherence to that dimension. For instance "Do you like music [Yes/No]" is part of the "interest-in-music" dimension. The model is measured to agree with its response exactly matches the particular response (room for improvement here to make the match less brittle!)
2. Using perspective API. Perspecitve API can measure [6 attributes](https://perspectiveapi.com/how-it-works/). We have set up the code necessary to use this api, assuming you have the appropriate api key in your environment variables. One issue with this approach to assessment is the RPM, which is 60 by default. This is perhaps an ok amount for evaluating usage of a product, but for evaluation it is way too slow! 
3. Using custom "assessmnent_functions", which take in text and return a score. Currently, no assessment functions are defined. We should add in a toxicity classifier at least.

All of these assessment approaches end up evaluating the output (or prompt/output pair) and aggregating the results for a particular dataset. Currently the anthropic evaluation interpretation is explicitly dependent on the dataset. The perspective api assessments are not labeled with the dataset, which is an issue and should be corrected. A toxicity measure of XXX means different things depending on the kind of prompt

## To Do
1. Dataset label should be added in some way to perspective (or assessment function) scores
2. Scores must be converted to evidence objects
3. Assessment functions needed to be provided and tested
4. prerun checks must be added
5. "Fairness" evaluations should be done (the prompt dataset can have a `sensitive feature` attribute, and evaluations should be disaggregated across it if it is defined)
6. Other forms of fairness should potentially be evaluated? For instance, the pronoun fill? 

- [ ] I have performed a self-review of my code

## Extra-mile Checklist


IanAtCredo:
Datasets should be downloadable rather than part of the package. 

Datasets that we are currently using:
1. [Bold](https://github.com/amazon-science/bold/tree/main/prompts) (These are currently slightly modified in our repo. Will have to refactor to use the original datasets)
2.[ anthropic evals](https://github.com/anthropics/evals)
3. [realtoxicity](https://allenai.org/data/real-toxicity-prompts) (also slightly modified, which seem unnecessary)

PR #319, opened 2023-03-16
Added support for ks score to classification models

fabrizio-credo:
## Added support for KS score for binary models

Duplicated ks_score metric for binary classifiers.

It's not super clean, but creating a single function incurs in other issues with the `metric_frame`.

I created aliases for the names:

```
"target_ks_statistic": ["ks_score_regression", "ks_score"],
"ks_score_binary": ["ks_score"]
```
so existing notebooks shouldn't break, but you can refer to the metric using `ks_score` and it automatically select the correct one.

@amrasekh if you have notebooks running ks_score for regression models, please test to make sure they aren't broken.

### DIFF
--- a/credoai/modules/constants_metrics.py
+++ b/credoai/modules/constants_metrics.py
@@ -16,6 +16,7 @@ from credoai.modules.metrics_credoai import (
     false_omission_rate,
     gini_coefficient_discriminatory,
     ks_statistic,
+    ks_statistic_binary,
     multiclass_confusion_metrics,
 )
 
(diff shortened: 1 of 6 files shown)


amrasekh (approved):
Looks good.
Let's make sure the [metrics doc page](https://credoai-lens.readthedocs.io/en/stable/pages/metrics.html#) also reflects that. Not caused by this, but the [Target_Ks_Statistic's](https://credoai-lens.readthedocs.io/en/stable/pages/metrics.html#target-ks-statistic) "Source: click here" currently lands on a `404 - page not found` and can be possibly fixed alongside too.

PR #328, opened 2023-04-07
Feat/pytorch support

esherman-credo:
## Describe your changes
Add support for pytorch models.
- Modify base_data to accept pytorch dataloader objects --> this could potentially work out of box for keras dataloaders, but unclear at the moment
- Modify classification model to support PyTorch models with HEAVY constraints -> user must specify some extra attributes in their custom-defined model class

https://credo-ai.atlassian.net/browse/DSP-475

- Need to confirm data validation is what we want it to be
- Comments

- [x] Will this be part of a major product update? If yes, please write one phrase about this update.
    - This adds support for PyTorch. It is very limited at this time (just classification models with essentially sequential architectures). Allows us to "play in" the PyTorch ecosystem.

## Extra-mile Checklist


credoai/artifacts/model/regression_model.py
```diff
@@ -1,6 +1,15 @@
 """Model artifact wrapping any regression model"""
 from .base_model import Model
 
+from utils.model_utils import reg_handle_torch
+
+try:
```
fabrizio-credo, line 6: As it is now this will be triggered anytime a user is using the artifact and has no `torch` installed. Can it be made specific to only attempts to handle torch models without the `torch` library?

credoai/utils/model_utils.py
```diff
@@ -125,3 +170,165 @@ def validate_dummy(model_like, _):
...
+    ):
+        predict_proba_obj = pred_func
+    else:
+        pass
+        # predict_proba is not valid (for now) --> this would correspond to multi-dimensional outputs or something similarly weird
+
+    return predict_obj, predict_proba_obj, clf_type
+
+
+def clf_handle_torch(model_like):
```
fabrizio-credo, line 223: I like how you separated handle functions. Can you add a docstring that gives the gist of what is going on?


```
### This is what Claude suggests 😛 
Creates a standardized interface for making inferences from a PyTorch classification model with different activations and output types.

- It checks that the model passed in has a forward method for making predictions. If not, it raises an error.
- It gets the output activation used by the model, either "sigmoid" or "softmax". If an invalid activation is specified, it defaults to "sigmoid".
- It performs a dummy prediction with the model to get the shape of the output. It uses this to determine if the model is a binary or multi-class classifier.
- It creates a predict function that performs inferences on either a single input or a batch of inputs. For sigmoid activation, it returns rounded predictions. For softmax, it returns the predicted class label.
- It creates a predict_proba function to return raw prediction probabilities. For sigmoid, it returns the logistic prediction. For softmax, it returns the probability of the positive class for binary classification or the full probability distribution for multi-class classification.
- It returns the predict and predict_proba functions along with a string specifying the classifier type (binary or multi-class).
- These returned functions can then be used to easily make predictions and get probabilities from the PyTorch model. The functions handle the differences between sigmoid and softmax activations and between binary and multi-class classification models.
"""
```

fabrizio-credo (changes requested):
All minor stuff, except the library import ones which for me are breaking.

Excellent work!

IanAtCredoAI (approved):
Looked through everything. The main change I made was making a PyTorch data class so that we don't make the Data class overly ugly.

Heroic work with the clf/reg handlers in model_utils. It's not the prettiest, but I was able to read through the web easily.

IanAtCredo:
Removed approval as I am building an example notebook. In that process it seems that my pytorch model is not being appropriately recognized (framework isn't being set correctly)

IanAtCredo:
Issues still outstanding:
* clf_torch_handle needs to define prediction functions that make use of the device used for training
* `output_activation` is janky. The fact that it needs to be specified isn't great. In addition, the implementation applies sigmoid/softmax to the output (as if the output was a linear layer). That's not necessarily the case.

